Issue #5, opened 2023-01-19
Add an online playground and initial project structure


adelhult:
For us to get started on the project and easily visualize the inner workings of the system it would be a great idea to add a web based playground (similar to [djot playground](https://djot.net/playground/)) that gets built automatically on every commit to the main branch. 
 
**Describe the solution you'd like**
Rust projects are often structured into multiple smaller disjoint library crates that then are sown together in another crate that imports them and provides an interface, this could for instance be a terminal app or bindings to an editor. 

For Modmark we could structure our repo into these main directories / crates:
```
parser (Takes text as input and produces a document tree)
compiler (Evaluates the document tree and hosts the wasm runtime. Might want to call this runtime/core instead...)
cli (Terminal tool that users will interact with)
playground (Allows us to inspect the document tree and play around with the parser and compiler)
```  

We should of course first and foremost focus on research and planning these first few weeks. But if we get this infrastructure set up we would have a good shared framework that we all can use to play around with our ideas.

**Describe alternatives you've considered**
We could just create a very simple terminal tool. But I think a website works better when we later down the line need to debug large large document trees.

adelhult:
I'll maybe throw together for this during the week-end, but would love to here if you thing it's a good idea or have another structure in mind.

CMDJojo:
Sounds good! I think it is a good idea to structure the repo into different crates. It makes sense to have one crate that is for the `cli` and one for the `playground` since they are two different wrappers around Modmark which has little to do with the language itself.

However, I think that we should have another structure for the other crates: a `core`/`runtime`/`modmark`-crate would be useful, which will be the one scheduling running the modules and such, keeping track of state and so on, but I do think we should have a `moduleloader`/`modules`/`modmgr`-module which finds, loads and keeps track of loaded modules. Since the data we will pass from/to a module will be an AST which will be transformed to another AST, with both of them being serializable (to json or something), there is nothing hindering us from letting `parser` have the same interface. Regardless if we will treat the `parser` as something configurable/modular like any other module, we can use the normal module interface for interacting with it in the first place. If we would decide that we should have special rule for it (which we most certainly will have in some form), we can just restrict the access compared to the other modules, which we might have to do for some other modules as well. My point is that I think the parser should be living along side the other standard modules and I don't see the need to create a special crate just for the parser itself; we will probably need a lot of standard modules (on the order of 20 is my best guess), so it would make sense to store the standard modules within whatever crate is managing them (probably the `modules` crate).

[...]

PR #6, opened 2023-01-19
Resolve GH-5: Add playground and initial project scaffolding

adelhult:
This PR including some basic scaffolding for the project which would allow us to get working on the different parts of the project without causing to many build issues for each other. I've included some "dummy" files but we will of course want to replace this with our code.

The project is structured into these crates:
* `cli` - terminal tool to transpile files
* `playground` - web based debug tool to inspect the document tree
* `core` - the crate that transpiles the document tree and hosts the wasm runtime
* `parser`  - takes a string and returns a document tree

As mentioned by @CMDJojo we would later want to add a crate or a directory of small crates containing all of the modules needed to craft actual documents too. This would be stuff like document templates and utility modules like `math`, `image` and  `code`. We might also later want to move the `parser` crate to these other modules that make up the system.

The PR also includes a web based "playground" intended to allow us to easily inspect the document tree and output as we work on the program. This may help us to easier play around with the program and share our progress with each other.

image


adelhult:
I think that this may be ready for review! : )

CMDJojo (approved):
Nice work! Great to have a playground up and running! OK to rebase into main

PR #7, opened 2023-01-22
Resolve GH-5: Improve playground and initial AST

CMDJojo:
This PR contains some improvements to the playground and the document tree first written by @adelhult , see GH-6 for the initial pull request. The changes were focused on making the initial tree look more like what might appear in the final product.

The project structure remains the same as implemented by @adelhult , however some internal restructurings were made: 

First of all, the `Element` enum now only has two cases, `Data(String)` and `Node{name: String, attributes: [String:String], children: [Element]}`, instead of the previous three, `Document(Box<Self>)`, `Paragraph(Box<Self>)` and `Text(String)`. Since the tree should be sent to and from different WASM-modules dynamically, it isn't possible to know the different possible nodes before-hand. Now, we have one leaf node, `Data(String)`, and one inode, `Node{..}`, which contains a string representing the type of that node. Later on, we might want to change the `Data` node to a node containing binary data, or to add a new leaf node containing binary data, or just keeping it as-is, and store binary data within external blobs and reference them from the tree. The `attributes` map is unused for now, but might store some state or configurations later on.

Second of all, a new string representation for `Element` was developed. The derived implementation would look something like this:
```
Node {
    name: "Paragraph",
    attributes: {},
    children: [
        Data(
            "Hej hej",
        ),
    ],
},
```
...the custom string representation would look like this:
```
Paragraph {
    attributes: { <empty> }
    children: [
        "Hej hej"
    ]
}
```

Since the `name` field is meant to signify different `Node` types, it improves readability a lot by having a custom string implementation.

The third thing that was added was two more output modes for the playground. In previous versions, there were two: `Document tree` and `Transpiled output`, but now there are four:

- Document tree (using the new custom string representation)
- Raw tree (using the one derived by rustc)
- HTML output (inspecing the HTML output without rendering it)
- Rendered HTML (renders the HTML output)

Additionally, some few changes were made in the wasm-bindings to accommodate the new features, in addition to a few changes in `core` to accommodate the merging of `Paragraph` and `Document` element (into the `Node` element).

This PR resolves GH-5

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -1,4 +1,5 @@
 use parser::Element;
+use std::collections::HashMap;
 
 // It is still somewhat unclear how we want to structure this part of the code base.
 // But basically we would transform the document tree by evaluating all of the nodes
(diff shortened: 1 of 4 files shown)


CMDJojo:
Moreover, compared to GH-6 which holds a placeholder parser which just wrapped the text within `Document { Paragraph { Data(text) } }`, this PR adds some more parsing capabilities: it now parses different lines into different text blocks and a blank line as a separator between paragraphs, similarly to how Markdown handles paragraphs. This is indeed just a placeholder implementation to show how the tree looks with multiple paragraphs and such.

adelhult (approved):
Looks good!

Issue #3, opened 2023-01-19
Add code style checks


CMDJojo:
It would be nice to have a rustfmt config so we all use the same formatting and add a CI tool that checks the style when doing a PR so we don't have badly formatted code in the main branch

axoxelol1:
I messed around with some github actions and found some things that work. However, I've only done it with a single cargo project in the repository so I am not sure what the best way to have multiple crates be tested is. Maybe just write more actions or if the project is a cargo workspace we might be able to run the commands on all everything  in the workspace. Unfortunately I do not know enough about rust project structures, or how we are structuring our project for that matter.

Here is an example that runs cargo check, cargo test, clippy, and rustfmt on pushes and PRs to main. If an action fails it will say what is wrong and how to fix it. https://gist.github.com/axoxelol1/786363790693f268838a798811b0a77d

adelhult:
Your gist looks great, good work!

Looks like if we put everything in a workspace then we would just be able to run the CI action once (at least for the test suite, don't know how well this works for clippy and fmt though?) 

We could also just repeat the command in each build but with different Cargo manifests specified
```yaml
with:
       command: check
       args: --manifest-path=./playground/Cargo.toml

with:
       command: check
       args: --manifest-path=./core/Cargo.toml

# etc ...
```

However, I think an issue with this is that if for instance the "playground" build fails, the job would stop and we would not get to see diagnostics for "core" or any other crate. So maybe we want to copy the entire action and just have multiple jobs like you suggested. I guess this might end up costing a lot of computational minutes though, since we would have to run `uses: actions-rs/toolchain@v1` for every every crate.  But I guess that we just have to try and see how expensive it ends up being :sweat_smile:.

I also took a look at this [action](https://github.com/rossjrw/pr-preview-action) that would allow us to host separate previews of the playground for each PR, but it might end up being a bit over-kill...

CMDJojo:
If I am not mistaken, you can use the `-p` option to run tasks on different packages, and I think you can specify it multiple times. It might work on sub-crates... In that way, we might be able to check formatting and tests on multiple crates on the same time.

CMDJojo:
But thank you for your gist! Will check it out tomorrow, and see if the `-p` option helps us out.

PR #8, opened 2023-01-22
Add CI for check/test/clippy/fmt

CMDJojo:
This PR attempts to use GitHub Actions as a CI for all pull requests and commits to main. To do this, two things have been done:
1. One Cargo Workspace has been added containing all four crates as members (and a `.gitignore` was added)
2. A `RustCI.yml` file was created under `.github/workflows` based on [this](https://github.com/modmark-org/modmark/issues/3#issuecomment-1398769696) comment from @axoxelol1 (some adjustments were made; names were added, toolchain installation was streamlined and more verbosity was added). But that gist was of good help!

The CI runs `check`, `test` , `clippy` and `fmt`.
I am unsure if this PR will actually check if the CI workflow works, but let's hope!

Resolves GH-3


axoxelol1:
We are getting some warning annotations when the CI runs due to actions depending on node 12.
>Code.js 12 actions are deprecated. Please update the following actions to use Node.js 16: actions/checkout@v2, actions-rs/toolchain@v1, actions-rs/cargo@v1. For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

Checkout could be updated to v3 to avoid this but I cannot find anything about updates to actions-rs. I didn't notice them when I made the gist. I think we could just ignore them for now and look at it again later.

CMDJojo:
Hmm, seems strange…didn’t notice those warnings at first. Thanks for the heads up! I can see what I find later today. Is there a good way to test-run the actions without pushing it to remote?Anyways, I think it is fine to ignore for now, and if anyone finds a solution, we can just reference this PR and the original issue from the new PR.For this to be merged, however, it would be nice if GH-7 gets approved first, since this builds on it.On 22 Jan 2023, at 20:20, axoxelol1 ***@***.***> wrote:﻿
We are getting some warning annotations when the CI runs due to actions depending on node 12.

Code.js 12 actions are deprecated. Please update the following actions to use Node.js 16: ***@***.***, ***@***.***, ***@***.*** For more information see: https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/.

Checkout could be updated to v3 to avoid this but I cannot find anything about updates to actions-rs. I didn't notice them when I made the gist. I think we could just ignore them for now and look at it again later.

CMDJojo:
@axoxelol1 regarding your previous comment:
> We are getting some warning annotations when the CI runs due to actions depending on node 12.

I have now updated checkout to use v3, and changed `actions-rs/*` to [`dtolnay/rust-toolchain`](https://github.com/dtolnay/rust-toolchain) and `run`. According to an [issue](https://github.com/actions-rs/cargo/issues/222) on the `actions-rs/cargo` task we previously used, that repo isn't maintained anymore. Thank you for the heads-up, would have missed the warnings otherwise! I cannot see any more warnings on the new run with the new CI, and if you think it looks okay, I think we are ready for merging!

CMDJojo:
We could also just `run` all the tasks sequentially and then not having to set up the toolchain one time for each job, but I think that it isn't worth it since the toolchain setup only takes 1s and accounts for ≈3% of the total runtime for most jobs, and with that little CI time, it is more worth to get the jobs done quicker in my opinion, but please tell me if you disagree!

axoxelol1 (approved):

axoxelol1 (approved):
lgtm

axoxelol1 (approved):
lgtm

Issue #14, opened 2023-01-27
Add wasm evaluator for modules


axoxelol1:
**Describe the solution you'd like**
There needs to be core functionality that can evaluate wasm modules.

**Additional context**
Will most likely use wasmer and we can hopefully get the evaluator running in a web browser.

PR #15, opened 2023-01-27
GH-14: Add a wasm runtime

adelhult:
This PR includes new code inside of `core` to load packages and evaluate module expressions. This is currently done with a `Context` struct. 

 We will deal with actually scheduling which order modules should be evaluated later. The main focus right now is actually getting to run and interopt with external packages.

## What WASM runtime will we use?
**We have opted to use [Wasmer](https://wasmer.io/) instead of [Wasmtime](https://wasmtime.dev/).** Wasmtime is more connected to the actual standards since it is developed by Bytecode Alliance. However, Wasmer seems slightly more document and also has support for compiling for the web, which mean that we might be able to get a online playground working :smile:.

## TODO before the PR is complete
- [x] Load and store packages, i.e. `context.load_package(...)` we might not need to care much about caching for now. We can deal with this in a later PR

- [x] Create an API to allow the package to tell the main runtime more about how it works, which version is it on, what nodes can it transform, what arguments does it accept?
   
   This can be done using functions that print information to stdout in the packages.
- [x] System for building standard packages to `wasm32-wasi` before importing them in core. We might want to use a `build.rs` which will let cargo do this for us.
- [x] Get wasmer to run on linux/windows/macos. We had trouble to get this to work when using a cargo workspace, the easiest way to solve this is by removing the workspace and rewriting our CI. But, it would be really nice if we could get this to work without removing the workspace.

   **Update**: This was fixed by adding `resolver="2"` in the workspace Cargo.toml file.

- [x] Get wasmer to run on the web with `wasm-pack build --target web`. We have gotten this to work, but not when using wasi. Might need to ask for help or open a issue in their repo.
   
   Looks like this will be possible with something like this:
   ```rs
   wasmer = {version = "3.1.1", default-features = false, features = ["js-default"]}
   wasmer-vfs = { version = "3.1.1", default-features = false, features = ["mem-fs"] }
   wasmer-wasi = {version = "3.1.1", default-features = false, features = ["js"]}
   ```
   But it will require quite a lot of conditional compilation since the web browsers built-in wasm compiler is used instead of cranelift when targeting the web.


bruhng (approved):
Very good job!

Issue #13, opened 2023-01-26
Add cli scaffolding with clap


hugomardbrink:
**Is your feature request related to a problem? Please describe.**
There should be a rudimentary way to interact with the compiler.

**Describe the solution you'd like**
The most rudimentary and accessible way is an CLI which should be implemented using clap. At this stage this should include some general info, version, debug and loading of files

PR #16, opened 2023-01-30
GH-13: Create cli base with watching.

hugomardbrink:
- Resolves: GH-13.

Uses clap for the cli, currently takes positional arguments for input and output file. Termion is used prettify and improve UX. This also implements a -w flag that supports compilation on save. Also implements -d for development mode which prints the tree on compilation

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,9 +1,9 @@
 use clap::Parser;
 use core::eval;
 use notify::{
-    Config, PollWatcher, RecommendedWatcher, RecursiveMode, Result, Watcher, WatcherKind,
+    Config, Event, PollWatcher, RecommendedWatcher, RecursiveMode, Result, Watcher, WatcherKind,
 };
-use parser::parse;
+use parser::{parse, Element};
 use std::env;
 use std::{fs, path::Path};
 use std::{fs::File, io::Write, time::Duration};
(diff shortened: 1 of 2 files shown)


cli/src/main.rs
```diff
@@ -1,16 +1,108 @@
...
+        print!("{}{}", termion::clear::All, termion::cursor::Goto(1, 1));
+        println!(
+            "{}Recompiling...{}\n",
+            color::Fg(color::Yellow),
+            style::Reset
+        );
+        match event {
+            Ok(_) => match compile_file(args) {
+                Ok(_) => {}
+                Err(_) => println!("[ERROR] File could not compile..."),
```
adelhult, line 60: Might be nice to use `eprintln` instead of `println` to print errors to stderr

cli/src/main.rs
```diff
@@ -1,16 +1,108 @@
...
+    #[arg(
+        short = 'w',
+        long = "watch",
+        help = "Watches file and compiles changes"
+    )]
+    w: bool,
+}
+
+fn compile_file(args: &Args) -> Result<()> {
+    let source = fs::read_to_string(&args.input).expect("[ERROR] Failed to read file...");
```
adelhult, line 30: Might be nicer to print to stderr this rather than panic:ing to get a prettier error message.

Perhaps we could use  `let... if` like this
```rs
let Ok(source) = fs::read_to_string(....) else {
   eprintln!("[ERRRO] ....");
};
```

adelhult (commented):
Looks great, I just found two small nit-picks

cli/src/main.rs
hugomardbrink: i opted to just return the errors instead

cli/src/main.rs
hugomardbrink: same thing with returning error here 👍

CMDJojo (approved):
Looks very good! Great job, this will be a nice addition to our project

Issue #29, opened 2023-02-07
Add syntax documentation


Forsinge:

Forsinge:
- [] Create a document describing functionality in detail.
- [] Create a reference document for default syntax

PR #37, opened 2023-02-07
Add parser test, migrate Element to Ast for IR of Parser, and prepare for supporting Tags

### LINKED ISSUE #22: Improve syntax
Issue #12 focused on making a "first draft for the syntax" and implement a parser following it, and PR #21 contained such a parser. The parser works fine for common scenarios, but lacks all features we want in the language. Additionally, although the PR contained some draft files for the syntax, it lacks a concrete definition for the syntax, explanation of the different constructions and what they do.

This issue focuses on continuing where #12 and #21 dropped of. More concretely, this issue focuses on:
- [ ] #29
- [ ] #36
  - [x] Deciding how they should work (wether they are customisable, what precedence they should have etc)
  - [x] Deciding their capabilities (if they only are to wrap some elements in a `Node` or if it should be a `ModuleInvocation` - if their content should be parsed already or if they are as free as any normal module invocation is)
  - [ ] Implementing it in the parser
- [x] Making a decision on if/how to escape from opening-bracket-environment (right now, `math(){(x + y + z)}` is impossible to write). Maybe if `!` is a part of the delimiter, it will be broken there and the `!` ignored?
- [x] Writing a test suite for the parser, with example modmark documents as input and expected tree structure as output
- [ ] #32
- [ ] Add actual error propagation to outer functions, instead of making a `Document` containing the text of the error
- [x] Maybe split syntax and semantics and thus turn `Element` to a `AbstractSyntaxTree`, removing `Environment` and maybe adding a `Tag` element

CMDJojo:
This pull request aims at implementing a lot of the issues mentioned in #22 . More concretely, here is the checklist and comments from #22 :
 - Add syntax documentation
   - This is tracked separately in #29 and is not included in this PR
 - Adding tags (like bold and italics in markdown)
   - This PR preserves more escaped characters during the first pass of the document, which will allow for parsing escaped tags
   - It also provides `Parser::second_pass`, which will be the function that will implement the tag matchings. It is currently just the identity function
   - The issue is tracked separately in #36 and the implementation will follow from that
 - Making a decision on if/how to escape from opening-bracket-environment, so `math(){(x + y + z)}` is expressible
   - We added rules for inline-modules to match just one character as opening-delimiter which will make this possible for inline modules
   - For multiline modules, they always start on a new row so that was already a non-problem
 - Write a test suite for the parser, with example input and expected parsedtree structure
   - `parser/tests` were added, with a file `compilation_test.rs` which takes either unified input-output files or split input modmark and output json files from `parser/tests/compilation_tests` and runs them as different tests
   - Additionally, the parser readme was updated with information on how the testing works and how to write your own
   - Moreover, a few example tests were added which will be expanded upon as more features gets added and more issues get identified and resolved
 - Add unrecoverable errors
   - This is tracked separately in #32 and is not included in this PR
 - Add error propagation
   - This relies on #32 and may be solved there or solved separately
 - Split syntax and semantics and add a new `AbstractSyntaxTree`
   - This PR includes an `Ast` struct which represents the syntax tree parsed by the parser. It can be turned to an `Element` struct, but doesn't include fields not related to the parsing such as `environment`, and has separate cases and structs for `Paragraph`, `Document` and `Tag` which is used to provide concrete types. All parser functions have been changed to use these types, up until the final `parse` which turns it into an `Element` using an `From` implementation

I think it would be great to rebase main onto this since the implemented features works as intended and there are already a couple of sub-issues related to this which requires this work for further development.

Resolves #22


adelhult (approved):
Looks great, good work! Really nice to be able to break the work on the parser into smaller subtasks now as well.

PR #39, opened 2023-02-08
GH-29: Add syntax documentation

Forsinge:
Resolves: GH-29

This adds a reference document for tags and smart punctuation in ModMark, as well as a description of its concepts and functionality. These documents will likely be updated later on to match changes and developments in the language.

### DIFF
--- a/MODMARK.md
+++ b/MODMARK.md
@@ -37,25 +37,23 @@ Example 2: paragraphs with escapes
 
 ## Modules
 
-Modules are invoked by an opening and closing square bracket. Information about the module is kept inside the square brackets, while the contents of the module is what follows. Exactly how much is consumed by the module depends on the closing delimiter, and whether the module is inline or multiline. For a module to be treated as multiline, it needs to be invoked outside other paragraphs, and without text on the same row as the square brackets.
+Modules are programs that are run from within your document, transforming the text you give them at their discretion. To use, or *invoke*, a module, encase its name within square brackets, and place the text you want to pass to it afterwards: `[url] https://example.com`. The module, in this case `url`, is invoked, consuming the text you give it, `https://example.com`.
 
-By default, an inline module will by default consume the next "chunk" of text, meaning the sequence of characters until a whitespace or newline character is found. This works the same even with a space immediately following the closing square bracket. A multiline module on defaults to consuming until an empty row is found.
+There are two ways to invoke a module, **inline** or **multiline**. When invoking a module in the middle of a paragraph, you do it **inline**. The module will consume the text following it, until the next space or until the end of the line, so you can write `Then, you take [math] x^2 and find out ...`. If you invoke the module **multiline**, you write it as if it was its own paragraph, with one empty row before and after it. It may then consume multiple rows. Here is one example of that:
 
-Delimiters are patterns specified after the closing square bracket. A module will then attempt to consume until the matching closing delimiter is found. However, inline modules are still limited to the row they were defined on, which means that they will close at the end of the row if no matching delimiter is found.
+```text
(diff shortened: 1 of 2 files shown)


CMDJojo:
Looks great!
One improvement I might suggest is adding a third column to the tags saying if they recursively parse their bodies for more tags or not (currently only Verbatim doesn't do that, so a note might be just fine). And maybe add Math as a tag as well?

Also, your reference made me notice that we don't have a tracking issue for headings as well, that's something we'll need to add. It isn't covered under #36 since it isn't actually tags per se. I will see what I can do tomorrow! But yeah, great work, it is looking good!

adelhult (approved):
I think this looks good! I guess we can go ahead and merge it right? 

We can always add more information once we have implemented headings and the modules are more complete.

CMDJojo (commented):
I think the document is kinda hard to follow, and I think that if we re-structure it a bit it will be much more true to the specification and easier to follow. I would like some changes before merging, as previousley discussed

CMDJojo:
I think you have gotten to a great start with the documentation, and I think it mainly looks good!

I would suggest changing the structure for some of the documentation in `MODMARK.md`. For example, the section about the modules is kind of split up - first, there is a wall of text describing what a module is in detail, followed by some examples further down, and then, at the end of the document, there is a list of what characters are allowed and not. I also think it is important to introduce terminology as we use it, so we don't say that something is restricted to its delimiters while not introducing what a delimiter is until two paragraphs further down.

I think sections like this will be easier to follow if the introduction of modules are in steps, with examples tightly integrated. For example, the structure may change from this:

## Modules
Modules are invoked by square brackets. Information about the module is kept within the brackets while the content is what follows. Exactly how much is consumed depends on the delimiters.
...
Example 3:
`[mymodule] This is a module`
Example 4:
`[mymodule]{&^ This text is included ^&}`
...
As module name, alphanumerics can be used. As delimiters, all special characters can be used.

My suggestion is to change it to something like this:
## Modules
Modules are programs that run on the text you give them, transforming the text at their discretion. To use, or invoke, a module, encase its name in square brackets and place its content following it: `[url] https://example.com`. The module, in this case `url`, is invoked, consuming its content `https://example.com` up until the end of the line or the next whitespace. ... To pass arguments to the module, configuring its behaviour in this specific instance, place them inside the square brackets: `[url color=pink] https://example.com`.

But you have done a great work! I think the content of the documentation is very good, and the `REFERENCE.md` is perfect, I just think that the content may be delivered in a different way to make it easier to read. If you disagree with my comment, feel free to let me know or merge this into main anyways - as Eli said, we can always make changes later!

CMDJojo:
To more clearly show what I mean, I have used your content and re-structured it a bit over at [cmdjojo/GH-29/0](https://github.com/modmark-org/modmark/tree/cmdjojo/GH-29/0), so that it to me looks clearer. Feel free to rebase those changes to this branch if you feel like they are good, or don't if you don't like them!

CMDJojo:
Sounds great! Even though I think some other sections may be restructured in the same way, this may be done later on. I think we should revise the part where \\ allegedly adds a hard break (it currently yields an escaped backslash), but other than that I think this may be merged. The documentation may be improved upon later on, if see fit. Good work, @Forsinge !

CMDJojo (approved):

CMDJojo:
Also, I do agree with you that the part with smart punctuation may be moved to the syntax specifications. You may move it there before merging, if you feel like it!

CMDJojo (approved):

CMDJojo:
It would be great to create a tracking issue for the smart punctuation when this is closed! Just so we don't forget it

PR #41, opened 2023-02-10
GH-32: Adds error handeling for wrong argument order

bruhng:
Resolves: GH-32

Adds error handling for unnamed arguments after named arguments. Right now it just panics if someone uses arguments wrong. But i am up for suggestions if someone has a better idea. The playground still works, the only difference is it does not update if there is a syntax error.

It is only the changes to the lib.rs file that is relevent. All other changes where just for my testing and will not be merged into main.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ Cargo.lock
 
 # MSVC Windows builds of rustc generate these, which store debugging information
 *.pdb
+
+.idea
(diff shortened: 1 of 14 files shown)


CMDJojo:
I suggest you rebase from main and push to this branch, so commits that have already been merged into main won't show up in this pull request. Also, when you say that `other changes where just for my testing and will not be merged into main`, it sounds like you aren't completley finished with this PR; there is a "draft pr" feature that makes it possible to work on the pull request while not being finished. That way, others will know that you aren't ready for merging, and doesn't spend time on a PR that might change a lot before actually being requested for merging.

CMDJojo:
For further reference: this PR has been reopened on a new branch, see #50

PR #42, opened 2023-02-12
GH-36: Add support for tags

### LINKED ISSUE #36: Adding tags (like  and  in markdown)

CMDJojo:
This PR handles #36 by adding support for tags. It also comes with several other improvements:
 * Multiple tests were added to test both non-stacked and stacked tags.
 * The module parsing, tag handling and general parsing is now split into three different files, `lib.rs` (the entry point/general parsing), `tag.rs` (extracting tags) and `module.rs` (parsing modules).
 * `Ast::tree_string`, which gives an output similar to `Element::tree_string(false)`, but cleaner and with the Element-specific things removed, to more clearly see the Ast content without things like Environment distracting.
 * Web bindings are added to have `Ast::tree_string` as an output option.
 * `Diffy` was added to show git-diff-like diffs for document compile tests, since seeing raw JSON makes it hard to compare.

Resolves #36


adelhult (approved):
Great work!

Issue #34, opened 2023-02-07
Inspect core::Context in playground


adelhult:
**Is your feature request related to a problem? Please describe.**
It would be very nice to be able to inspect the context and info about packages inside of the playground

**Describe the solution you'd like**
I will implement another tab inside of the playground to inspect the Context. 

**Describe alternatives you've considered**
-

**Additional context**
-

PR #43, opened 2023-02-13
GH-34: Add context inspector + better styling

adelhult:
This PR resolves GH-34 by adding an extra tab where the context (and the packages) can be inspected (see screen-shot).

The styling of the playground is also improved by adding nicer buttons, a better layout for phone sized screens and a few icons here and there.

image


axoxelol1 (approved):
Looks good, when approving I noticed it does not work in chrome, created an issue (#45) for this. However it is also broken in main so you did not cause this.

PR #50, opened 2023-02-13
GH-32: Add error checking for unnamed arguments after named arguments

bruhng:
Resolves: GH-32

Adds the enum MaybeArgs instead of Modulearguments to be able to handle when no valid arguments exists.

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -41,6 +41,18 @@ pub struct ModuleArguments {
     pub named: Option<HashMap<String, String>>,
 }
 
+#[derive(Clone, Debug, PartialEq)]
+pub enum MaybeArgs {
+    ModuleArguments(ModuleArguments),
+    Error(String),
+}
+
+impl Default for MaybeArgs {
+    fn default() -> Self {
+        MaybeArgs::ModuleArguments(ModuleArguments::default())
+    }
+}
+
 /// This enum represents an Ast, an Abstract Syntax Tree. It is essentially a tree-like structure
 /// representing the structure and content of a parsed document. `Text` and `Module` are leaf-nodes;
 /// they do not contain any other nodes, and all other nodes are inner nodes (they may contain
@@ -93,12 +105,16 @@ impl From<Ast> for Element {
                 environment: HashMap::new(),
                 children: tag.elements.into_iter().map(|e| e.into()).collect(),
             },
-            Ast::Module(module) => ModuleInvocation {
-                name: module.name,
-                args: module.args,
-                body: module.body,
-                one_line: module.one_line,
-            },
+            Ast::Module(module) => {
+                match module.args {
+                    MaybeArgs::ModuleArguments(args) => ModuleInvocation {
+                        name: module.name,
+                        args,
+                        body: module.body,
+                        one_line: module.one_line,
+                    },
+                    MaybeArgs::Error(error) => panic!("{}", error),
+                }},
         }
     }
 }
@@ -122,7 +138,7 @@ pub struct Document {
 #[derive(Clone, Debug, PartialEq)]
 pub struct Module {
     pub name: String,
-    pub args: ModuleArguments,
+    pub args: MaybeArgs,
     pub body: String,
     pub one_line: bool,
 }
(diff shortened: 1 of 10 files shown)


parser/src/lib.rs
```diff
@@ -93,12 +105,16 @@ impl From<Ast> for Element {
...
-            },
+            Ast::Module(module) => {
+                match module.args {
+                    MaybeArgs::ModuleArguments(args) => ModuleInvocation {
+                        name: module.name,
+                        args,
+                        body: module.body,
+                        one_line: module.one_line,
+                    },
+                    MaybeArgs::Error(error) => panic!("{}", error),
```
CMDJojo: I think that this shouldn't panic. This means that if one user writes any invalid arguments, the program just crashes. Try changing the trait from `From<T>` to `TryFrom<T>` to allow for one success type and one failure type, rather than panicking!

CMDJojo (changes requested):

CMDJojo:
I would also love some tests confirming that this works. Since we haven't implemented panic tests yet, you may skip that, but test some module invocations with correct arguments and see that they work. You can do that by looking at the test files already there and copy-paste that format. Then, maybe add some tests using different formats, for example:
1. Invocations without arguments
2. Invocations with unnamed arguments
3. Invocations with named arguments
4. Invocations with unnamed and then named arguments
This won't verify that the code crashes on malformatted arguments, but it will verify that it correctly parses correct arguments and doesn't crash in that case. Test your test cases with `cargo test`

parser/src/lib.rs
bruhng: This would require a lot of rewriting because we can not use into on any instance of Ast but instead try_into which would require checking for Result everywhere but maybe it is worth it.

CMDJojo: I do not think we need to do so much rewriting. If we do TryFrom, just wrap every case you currently have into an `Ok(...)` and the panicing case into a `Err(...)`. Note that this is just the final step when converting to Element, so no `Result`s will be in our Ast. Also, note chaining errors with the question mark operator. See the example on the `TryFrom` doc page on `GreaterThanZero(i32)` for an example

bruhng: image

```the trait bound `Element: From<Ast>` is not satisfied
required for `Ast` to implement `std::convert::Into<Element>`rustc[Click for full compiler diagnostic](rust-analyzer-diagnostics-view:/diagnostic%20message%20%5B2%5D?2#file%3A%2F%2F%2Fhome%2Fbruhng%2FGitRepos%2Fmodmark%2Fparser%2Fsrc%2Flib.rs)
lib.rs(104, 68): required by a bound introduced by this call```

This is because we do not implement into anymore but try_into. But maybe we do not use it in that many places.

Comment (CMDJojo) [reply parser/src/lib.rs]: We just need to change our .into() to .try_into(), I don't think it will be a problem

Comment (bruhng) [reply parser/src/lib.rs]: But then how do we handle try into when it fails? Do we panic? We are back where we started

Comment (bruhng) [reply parser/src/lib.rs]: I don't know the best way to handla it either so I am open to suggestions. If the playground panics it just doesn't update untill the text is fixed, it does not crash, if that helps our decision, we can discuss it tomorrow or Thursday

Comment (CMDJojo) [reply parser/src/lib.rs]: I do not think the code should panic at all. The best way, in my opinion, would be to propagate it to callee, as written in the original issue. The callee would need to handle the unwrapping and what to do if something fails, which is hopefully not panicking, but rather showing an error message or something. And yes, the HTML doesn't crash in the playground, but the panic makes the entire rust-part of the playground crash, which would lead to bad things if we would have persistant state which may not be recoverable, or loaded/cached modules which would be lost. One way would be to try to catch the panic and unwind - which would lose all diagnostics and add overhead. The other option is to use `TryFrom` instead, which I think works fine. The rust docs also says that this is the way to do it - `From` may not fail but `TryFrom` may return an error. I still strongly suggest we use that and don't push panicking code to main. If we run the program with the watcher from cli, it would actually completly crash

**Comment by bruhng** [comment]
Added the 4 tests that where requested 

> I would also love some tests confirming that this works. Since we haven't implemented panic tests yet, you may skip that, but test some module invocations with correct arguments and see that they work. You can do that by looking at the test files already there and copy-paste that format. Then, maybe add some tests using different formats, for example:
> 
> 1. Invocations without arguments
> 2. Invocations with unnamed arguments
> 3. Invocations with named arguments
> 4. Invocations with unnamed and then named arguments
> 
> This won't verify that the code crashes on malformatted arguments, but it will verify that it correctly parses correct arguments and doesn't crash in that case. Test your test cases with `cargo test`

and they all pass:
image

**Comment by CMDJojo** [comment]
I have currently been unable to review the code, but I think I will be able to do that tomorrow morning!

File: parser/src/lib.rs [hunk 2]
```diff
@@ -74,30 +86,47 @@ impl Ast {
...
+                    .collect::<Result<Vec<Element>, String>>()?,
+            }),
+            Ast::Module(module) => match module.args {
+                MaybeArgs::ModuleArguments(args) => Ok(ModuleInvocation {
+                    name: module.name,
+                    args,
+                    body: module.body,
+                    one_line: module.one_line,
+                }),
+                MaybeArgs::Error(error) => Err(error),
```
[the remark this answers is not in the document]
CMDJojo: Also, since the error is always the same, the type may be `&'static str` (I think), to avoid unnecessary copying.

parser/src/module.rs
```diff
@@ -266,11 +266,28 @@ fn parse_module_name(input: &str) -> IResult<&str, &str> {
...
+fn get_module_args_parser<'a>(inline: bool) -> impl Parser<&'a str, MaybeArgs, Error<&'a str>> {
     map(
         opt(alt((
+            map(
+                tuple((
+                    separated_list0(
+                        get_arg_separator_parser(inline),
+                        get_unnamed_arg_parser(inline),
+                    ),
+                    many0(get_arg_separator_parser(inline)),
```
CMDJojo, line 278: I don't think this `many0` is needed since `get_arg_separator_parser` is `space1`/`multispace1` and thus catches as many spaces as possible (we can't reapply it when it already caught all spaces)

CMDJojo (changes requested):
Nice tests and such! The code looks much better now! However, there are a few places where `panic!`s are still used when it can actually crash the program which I would love to have revised before merging. Also, some malformed args like `[foo bar=baz cux]` is caught, but `[foo bar=baz cux=]` isn't, is this intentional or are we not trying to catch those?

CMDJojo (approved):
Looks good! Great work! Ready for merge

Issue #44, opened 2023-02-13
Add headings


Forsinge:
**Is your feature request related to a problem? Please describe.**
There is currently no support for headings, and it is not included in #36 since tags and headings function differently.

**Describe the solution you'd like**
Correct parsing of headings of various levels, likely 1 to 6 like HTML and Markdown.

PR #56, opened 2023-02-16
GH-44: Add headings

Forsinge:
Resolves: #44 

Adds headings in the parser. This allows syntax like ``# Title`` to be parsed as a "Heading1" node.

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -138,11 +138,15 @@ impl TryFrom<Ast> for Element {
                 }),
                 MaybeArgs::Error(error) => Err(error),
             },
-            Ast::Heading(heading) => Node {
+            Ast::Heading(heading) => Ok(Node {
                 name: format!("Heading{}", heading.level),
                 environment: HashMap::new(),
-                children: heading.elements.into_iter().map(|e| e.into()).collect(),
-            },
+                children: heading
+                    .elements
+                    .into_iter()
+                    .map(|e| e.try_into())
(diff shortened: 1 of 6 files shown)


parser/src/lib.rs
```diff
@@ -259,13 +272,46 @@ fn parse_document_blocks(input: &str) -> IResult<&str, Vec<Ast>> {
...
+///
+/// # Arguments
+///
+/// * `input`: The text to parse
+///
+/// returns: The heading node, if a successful parse occurs, otherwise the parse error
+fn parse_heading(input: &str) -> IResult<&str, Heading> {
+    map(
+        pair(
+            verify(take_while1(|c| c == '#'), |s: &str| s.len() <= 6),
```
adelhult, line 298: I think we should allow an arbitrary amount of levels for headings (limited of course by the data type used, let's say an u8) and then each "Heading" transforms can be responsible for the how to handle the level.

This would avoid being too HTML centric, in LaTeX for instance there might not be a meaningful interpretation of Heading 6 since it stops at `\subsubsection{...}`. And some other presentational format might find it handy to support even more levels.

adelhult (commented):
Looks really good! Good luck with the build errors!

(Side note, I might add an arguments field to Element::Node in my refactor of core. This means that we could use just one "Heading" and assign different attributes for the levels. But that can of course be added after this PR is merged!)

adelhult (approved):
Great work, thanks for the change too.

Issue #35, opened 2023-02-07
Add transform and evaluation functionality in core


axoxelol1:
**Is your feature request related to a problem? Please describe.**
Now that there is a system for loading packages the next step is to implement functionality for transforming and evaluating module blocks. This means implementing `eval` in `core/src/lib.rs` and `transform` in `core/src/context.rs` or something equivalent. 

**Describe the solution you'd like**
For now any form of evaluation that works with basic documents is fine, some bold and italic. Thinking about scheduling and getting table of contents/citations to work will probably be in a future PR.

PR #64, opened 2023-02-16
Evaluate documents

adelhult:
Resolve GH-35. This PR adds a `eval` and `transform` function as well as general improvements to the playground. This makes it possible to evaluate entire documents.

This PR got really big, and is far from perfect but I think it would be nice to go ahead and rebase it into main and make smaller issues for everything needed to be fixed. But feel free to take a look and give input if you have any questions or find errors.

TODO:
- [x] Add `content_inline` and `content_block` modules
- [x]  Refactor Element and rebase with main (blocked by #50, and #56)
- [x] Document how the module API works
- [x] Clean-up, would be nice to remove a lot of needless clones
- [x] Fix broken arg handling and add args to Element::parent
- [ ] ~~Change name conventions for modules~~ I'll leave this as is and let us discuss it on thursday together.
- [x]  Fix broken packages
- [x] Fix multiline json serialize bug


Cargo.toml
```diff
@@ -5,7 +5,7 @@ members = [
     "parser",
 ]
 
-exclude = ["modules/test-module", "playground/web_bindings"]
+exclude = ["modules/test-module", "modules/table", "modules/html", "playground/web_bindings"]
```
CMDJojo, line 8: Do you think wildcards are allowed, so we can do `modules/*`? Would be nice so we don't miss adding a module

axoxelol1 (approved):
Works well on my machine. Good job getting the last stuff done, exciting to finally see some rendered documents. There is a memory leak issue (#80) however I still think this should be rebased to main to make developing packages easier.

cli/src/main.rs
```diff
@@ -25,6 +25,14 @@ struct Args {
     #[arg(index = 2, help = "Path to output file")]
     output: String,
 
+    #[arg(
```
CMDJojo, line 28: Nice change adding a flag! Will help if we have multiple output files (i.e if one module creates image files that are included later on etc). Do you think it is worth comparing this flag value with the extension? Now, some may output the file `output.latex` and expect it to be in latex format rather than html. Might be added later on

cli/src/main.rs
```diff
@@ -36,21 +44,21 @@ struct Args {
...
-    let output = eval(&document, &mut ctx);
+    let output =
+        eval(&source, &mut ctx, &OutputFormat::new(&args.format)).expect("Failed to evaluate file");
 
     let mut output_file = File::create(&args.output)?;
     output_file.write_all(output.as_bytes())?;
 
-    Ok(document)
+    // Also return the Element tree for debug purposes
+    Ok(parse(&source).unwrap())
```
CMDJojo, line 57: Shouldn't this be `parse(&source)?` to propagate errors rather than crashing? Or is the panic caught somewhere else? Because, if I understand it correctly, the CLI will crash on parser errors

core/src/context.rs
```diff
@@ -1,42 +1,50 @@
...
+            transforms: HashMap::new(),
             store: get_new_store(),
         }
     }
 
     pub fn load_default_packages(&mut self) {
         self.load_package(include_bytes!(concat!(
             env!("OUT_DIR"),
-            "/test-module/wasm32-wasi/release/test-module.wasm"
+            "/table/wasm32-wasi/release/table.wasm"
```
CMDJojo, line 34: Seems like a fine way to do it for now, but it would maybe be nice to make a macro for loading to decrease repetition

core/src/context.rs
```diff
@@ -48,41 +56,304 @@ impl Context {
...
+    pub fn deserialize_compound(input: &str) -> Result<Element, CoreError> {
+        let entries: Vec<JsonEntry> = serde_json::from_str(input)?;
+
+        // Convert the parsed entries into real Elements
+        let elements: Vec<Element> = entries.into_iter().map(Self::entry_to_element).collect();
+        Ok(Element::Compound(elements))
+    }
+
+    /// Convert an JsonEntry to a Element
+    fn entry_to_element(entry: JsonEntry) -> Element {
```
CMDJojo, line 186: It might be worth to `impl Into` here, maybe? It will also be clearer that this is associated with `Element`

core/src/lib.rs
```diff
@@ -1,64 +1,145 @@
...
+            args: _,
+            ref body,
+            inline: _,
+        } => {
+            // Base case, if its just raw content, stop.
+            if name == "raw" {
+                return Ok(body.clone());
+            }
+
+            if name == "inline_content" {
```
CMDJojo, line 94: This isn't something that needs do be done now, but I think we would want to move the "native" modules out to a separate file since we will need to add some more later on, and add some proper manifests to them. It would be nice if we could use the same `Module` struct as before, and have an enum for the implementation, if it is implemented in wasm or natively by us, to make it more uniform

playground/static/index.html
```diff
@@ -66,17 +67,64 @@
...
-                    pre.innerText = ast(input)
-                    break;
-                case "transpile":
-                    pre.innerText = transpile(input);
-                    break;
-                case "render":
-                    output.innerHTML = transpile(input);
-                    break;
+                case "ast": {
+                    let result = "";
```
CMDJojo, line 77: Another improvement would be to move the common parts out, `result`, `try-catch` and so on

CMDJojo (approved):
Looks really good! Fantastic work! I had some comments, you may want to look through them, there are not so much things we need to fix now-now but if you feel like some of the things are good to fix before merging, go ahead!

cli/src/main.rs
adelhult: Good idea I created GH-81.

Cargo.toml
adelhult: Also a nice idea, tried it out but it does not seem to work. Can try a bit more once we refactor the modules directory.

core/src/context.rs
adelhult: Agree! But I think we can add that later : )

core/src/context.rs
adelhult: Yeah, I wondered about that too. However,  since `element_to_entry` requires additional information it could not be implemented for that function and I thought it be best to keep the other way around as function too to avoid confusion.

core/src/lib.rs
adelhult: Good suggestion, I've created GH-82.

playground/static/index.html
adelhult: Fixed!

cli/src/main.rs
adelhult: Fixed!

PR #65, opened 2023-02-16
Fix deployment and clean up error handling

CMDJojo:
This is a small increment to #50 and does the following:
 1. Clarify the error message that appears on Nom errors (which normally shouldn't appear at all, but must be handled anyways)
 2. Removes `Element` views from the playground; they aren't that useful and is a struggle to maintain. The `Ast` view may represent the same data as the `Element` does and more. The `Ast` view may show errors while the `Element` would fail and need a fall-back or force-unwrapping which are both bad

Resolves GH-32

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -99,7 +99,7 @@ impl TryFrom<Ast> for Element {
 
     fn try_from(value: Ast) -> Result<Self, Self::Error> {
         match value {
-            Ast::Text(s) => Ok(Data(s)),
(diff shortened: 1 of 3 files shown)


CMDJojo:
Note that the previous playground failed to deploy since there were problems in the web bindings which weren't properly checked. This will build the playground properly, but force-unwraps the document when transpiling. This will change when `core` is updated and allows error handling

bruhng (approved):
Looks great!

PR #67, opened 2023-02-16
GH-66: Add previews to PRs

CMDJojo:
This PR aims to provide a preview for the playground on each pull request. This included changing the pages from being based on just artifact uploads (as it was done previously) to be based on a new branch `gh-pages`. The history of this branch will be cleaned at every push to `main`, so there is no risk that the history will be huge.

Resolves GH-66

PR #68, opened 2023-02-16
Example PR for testing preview playgrounds on pushes to main

CMDJojo:
This is a "dummy" draft PR that is opened to see that when #67 is merged, this playground preview still is live. This is not intended for being merged, and will be closed as soon as we see the result of this test.


CMDJojo:
The test was a success, since the deployment was active even when a push to main was done. This PR will thus be closed

Issue #66, opened 2023-02-16
Preview playground on PR:s


CMDJojo:
**Is your feature request related to a problem? Please describe.**
Currently, when making PR:s, CI checks formatting, tests and some code correctness, but it currently doesn't check the web bindings and deploymentability. If a branch fails to deploy, it may thus not fail any checks until it is merged into main and attempted at deployment.

**Describe the solution you'd like**
I suggest we add the ability to preview playground deployments on pull requests. This will
 1. Ensure that the branches are actually possible to deploy
 2. Give us a good look at how it would look like before merging to main
This needs us to change from deploying a single artifact with gh-actions, and change our workflow to the more traditional branch-based page deployment. It will probably be a good idea to base it on [pr-preview-action](https://github.com/rossjrw/pr-preview-action)

**Describe alternatives you've considered**
One alternative is to add a CI step running the deployment steps but without deploying the page, which is run on every PR to check if the deployment compiles. This will catch build errors but not provide any preview.

PR #69, opened 2023-02-16
GH-66: Do not build pages when pull request is closed

CMDJojo:
This PR changes `DeployPreview.yml` to not build the page if the pull request is closed. It is still important that the workflow runs, since it needs to do that to remove the preview playground page when the PR is closed.


bruhng (approved):
Looks good!

Issue #55, opened 2023-02-15
Add smart punctuation


Forsinge:
**Is your feature request related to a problem? Please describe.**
Now that tags have been implemented and headings are being worked on, it seems like a natural next step to extend parser functionality with smart punctuation.

**Describe the solution you'd like**
Automatic replacement of characters to achieve what was specified in #29.

PR #72, opened 2023-02-17
Add smart punctuation

Forsinge:
Resolves: #55 

This PR adds smart punctuation.

TODO:
- [x] Fix smart quotes with elements between
- [x] Fix newlines to pass tests
- [x] Allow escaping quotes and sequences
- [x] Avoid replacing longer sequences such as "----"
- [x] Avoid using single quotes in words like "can't", possibly by using left- and right-flanking quotes
- [x] Bugfixes: escaped quotes like ``"one \" two"`` and single quotes like ``'one two t'hree four'``
- [x] Clean-up

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -399,69 +399,88 @@ fn smart_punctuate<T>(input: &mut T)
 where
     T: CompoundAST,
 {
-    let mut sgl_open = false;
-    let mut dbl_open = false;
+    let mut backtrack: Vec<(usize, bool)> = vec![]; // bool indicates dbl or sgl
+    let mut sgl_open: Option<usize> = None;
+    let mut dbl_open: Option<usize> = None;
 
-    input.elements_mut().iter_mut().for_each(|e| match e {
-        Text(str) => {
-            let mut acc = String::new();
-            let mut lines = str.lines().peekable();
+    input
+        .elements_mut()
+        .iter_mut()
+        .enumerate()
+        .for_each(|(i, e)| match e {
+            Text(str) => {
+                let mut acc = String::new();
+                let mut lines = str.lines().peekable();
 
-            while let Some(line) = lines.next() {
-                let mut tmp = line
-                    .replace("...", "\u{2026}")
-                    .replace("---", "\u{2014}")
-                    .replace("--", "\u{2013}");
+                while let Some(line) = lines.next() {
+                    let mut tmp = line
+                        .replace("...", "\u{2026}")
+                        .replace("---", "\u{2014}")
+                        .replace("--", "\u{2013}");
 
+                    if let Some(ii) = &sgl_open {
+                        tmp = tmp.replacen("\'", "\u{2019}", 1);
+                        backtrack.push((*ii, false));
+                    }
+                    if let Some(ii) = &dbl_open {
+                        tmp = tmp.replacen("\"", "\u{201D}", 1);
+                        backtrack.push((*ii, true));
+                    }
 
-                if sgl_open {
-                    tmp = tmp.replacen("\'", "\u{2019}", 1)
+                    let sgl_occs = tmp.matches("\'").count();
+                    for _ in 0..sgl_occs / 2 {
+                        tmp = tmp
+                            .replacen("\'", "\u{2018}", 1)
+                            .replacen("\'", "\u{2019}", 1);
+                    }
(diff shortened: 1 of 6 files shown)


CMDJojo:
Really great work! I see that you have put lots of effort into this, and made some smart design choices, like not treating the `'` within words as a smart punctuation candidate.

When playing around, I noted some discrepancies between how I assumed escapes would work and how they actually worked (not since they are wrong, not at all, just since I wrote the tag parsing and I handled them differently there). As I see it, we have two options for how escaping smart sequences should work:

## Option 1
The first option is that a backslash `\` escapes the **next following sequence**. `\...` will not only escape the first dot but all of them.
### Pros:
With this implementation, an emdash may be escaped with just one backslash: `\---` will result in three separate dashes. It also makes it easy to escape long sequences of dots, like `\...........`
### Cons:
This makes certain combinations of sequences impossible to write, and non-symmetic by design. `\....` will equate to for dots while `...\.` will equate to one ellipsis and one dot. Also, any sequence of the same characters where you want it followed by a smart sequence of that character is simply impossible to express, you can't write a dot followed by an ellipsis, since `\....` doesn't work and `\. ...` adds a space inbetween, or a dash followed by an endash or emdash. It is also inconsistent with non-sequence-based smart punctuation, `\....abc...` doesn't surround `abc` in ellipsis, but `\""abc"` surrounds `abc` in smart quotes.

## Option 2
The second option is that a backslash `\` escapes **the next character** and doesn't allow it to be a part of a smart sequence. `\...` will only escape the first dot, and the resulting two dots aren't enough to be a part of the ellipsis.
### Pros:
All sequences of characters including smart sequences are allowed to be expressed in the language. You may write a dot followed by an ellipsis by doing `\....`. It also allows for dashes followed by en/emdashes and such. The syntax is symmetric since it always escapes just one character, both `\....` and `...\.` gives an ellipsis next to a dot.
### Cons:
If users assume that it will escape any sequence, it will not work as expected from them. You can't escape an en/emdash with one backslash, `\---` yields a dash and an endash, while you need `\-\--` to fully escape both the emdash and endash. Also, if you want long sequences of dots, you'll need to add a backslash between (at least) a third of them: `\...\...\...`.

As the time of writing this, the semantics of **Option 1** is implemented by this PR while **Option 2** is used for tag parsing. You may do `**abc\***` to get `<b>abc*</b>` (or equivalent in other languages). We need to decide if we are going to use **Option 1** or **Option 2** for smart punctuation. It comes down to usability; is it more useful to be able to write a dash followed by an endash, or is it worth sacrificing the ability to do that to more easily escape emdashes?

If we go for **Option 1**, we have another important consideration; should we change the tag parser or not? If we do not change the parser, the language may feel inconsistent, `...abc\....` will not encase `abc` in ellipsis but `**abc\***` will make `abc*` bold. If we change the tag parser, we will get the same problems expressed in **Cons** under **Option 1** but maybe even worse since bold stars and verbatim backticks are arguably even more usefull by a dot-preceeding ellipsis.

Personally, I argue for **Option 2**. I do think that sequences like `.......` are seldom wanted and the extra escaping needed won't be much of an issue, while I strongly think that it is important to be able to express corner-cases like an dash followed by an emdash.

Also, I personally don't think changing the tag parser is good at all. `\`this\`` in a verbatim environment won't be expressible, and it isn't hard to find a case where you may want a star at the end of a bold sentence, like if you want to add a note to some bold text indicated by a trailing star. So I really think the tag parser should stay at **Option 2** regardless, and keeping the smart sequence parser at **Option 1** will be an inconsistency I think is bad. Additionally, it will be easy to explain that **backslashes escapes the following character so that it may not be parsed as being part of a tag, smart sequence or any other smart punctuation character**, and I think it is easier to get an intuition for that, than having different semantics in different places; **backslashes before a character that is not a smart punctuation sequence such as ellipsis, endash and emdash escapes all of the following identical characters, while backslashes before any other character, part of possibly a tag or a smart punctuation pair (not sequence) will only escape that character** is in my opinion much more of a struggle to learn.

I would love some input on these thought! Tell me if you think we should change the tag parser, the smart punctuation parser, both or none, and how! And if you have any other options or suggestions how escapes may be treated, let me know.

CMDJojo (approved):
Looks good! The code is quite complicated, but that's what happens when you do string/char manipulation. But good job!

PR #75, opened 2023-02-17
GH-74: Build modules async

### LINKED ISSUE #74: Make modules build async
There is a build script `Build.rs` in `core` that builds all wasm modules in the `modules` directory. Currently, this is done sequentially, so that the first module is built first, then the second one, then the third one etc. This slows down the builds (especially for CI)

To build all the modules in parallell

CMDJojo:
This PR changes the `Command` from using `.output()` (which blocks the thread until the program is finished) to using `.spawn()` (which spawns the process without blocking), and then loops through all spawned threads and waits. This builds the modules in parallel.

Resolves #74

### DIFF
--- a/core/build.rs
+++ b/core/build.rs
@@ -1,3 +1,4 @@
+use std::process::Child;
 use std::{
     env, fs,
     path::{Path, PathBuf},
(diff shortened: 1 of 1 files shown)


adelhult (approved):
Looks good! It worked fine to compile it on my machine too.

PR #77, opened 2023-02-20
GH-22: Use macros in or.rs to avoid repeated code

CMDJojo:
`or.rs` is a file which allows combining different parsers of different types in a way where all parsers are tested and the result of the first one which succeeds is returned. It added functions `or2` `or3` ... `or7` to support different amount of parsers, and even though the functions were simply delegating work to other functions, it was tedious and error-prone work to write ~55 lines for just one instance, where one mistake may make the result completely invalid.

This PR replaces the old hard-coded 6 functions by functions generated by macros. Now, there are `or`-functions for any amount of parsers from 2 to 26. I think this is much better since the automatic generation makes sure that there can't be mistakes due to carelessness when copy-pasting the function to add one argument, and since the parsing tests succeeds (which are using these new functions instead of the old functions), I am certain that they work well.

There are comments attached to all macros, but if you want to see them expanded, you can use `cargo-expand` to expand them (but beware, since there are so many functions, there are over 5000 lines generated.

Even though this may seem "overkill", many of the functions in `nom` itself is macro-generated for many different number of arguments. `rustc` makes sure that no unused code is moved into the binary, so I really do not see any problem using this.

### DIFF
--- a/parser/Cargo.toml
+++ b/parser/Cargo.toml
@@ -8,6 +8,7 @@ edition = "2021"
 [dependencies]
 nom = "7.1.3"
 thiserror = "1.0.38"
+paste = "1.0.11"
 
 [dev-dependencies]
 json = "0.12.4"
--- a/parser/src/or.rs
+++ b/parser/src/or.rs
@@ -13,7 +13,8 @@
 //! of the parsers match, the or combinator fails.
 //!
 //! The functions are called `or{n}` with `n` being the amount of parsers to be combined, currently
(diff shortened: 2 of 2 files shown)


CMDJojo:
I decided to remove some of the `or` functions since it seems to slow down build times targeting the web for some unknown reason. There are now `or2`-`or10`

bruhng (approved):
Great work, makes the code more readable and less bloated.

Issue #80, opened 2023-02-20
Rendered html output runs out of memory


axoxelol1:
**Describe the bug**
When using the rendered html output (introduced in PR #64) the playground runs out of memory after using it for a while. Could possibly try to fix this at the same time as #45 and #78 if you were going to take a look at those two anyway @adelhult.

**To Reproduce**
Go to the website and start writing multiple modules, headings etc. I am using firefox which might be relevant.

**Screenshots**
![image](https://user-images.githubusercontent.com/40837400/220162785-e2f5dae1-0f49-4bf0-a072-b483b295926e.png)

axoxelol1:
Hm I supposed we could. It would be nice to merge #64 to make it easier to work on new packages but I agree that it is not optimal. I could have a look and see if it is anything obvious but I'm afraid it might be hard to track down the cause.

adelhult:
I think it might be nice to merge it anyway since we won't continue blocking people from starting writing modules.  And it only causes trouble when using the eval function and not when testing the parser.

adelhult:
I created a first attempt at fixing the problem, but sadly I've reached a dead-end. I thought that I could solve both this and #45 by serializing and deseralizing the modules everytime they are used to avoid having a shared Engine and Store. But, sadly it seems like it's impossible to deseralizing on the web since the methods only exists and (make sense) when we have the "compiler" feature flag enabled.

This is looking quite bad, especially for #45. But I think both of the issues still can be fixed with a bit of luck.

I'll now attempt at  having a shared Engine in the Context object and new stores each time a module is invoked. The problem with this approach is that stores seem to hold a reference to the engine and I wonder if this will have any unwanted side effects. But hopefully it will work fine.

PR #83, opened 2023-02-21
Unify structure for external, native and built-in packages

### LINKED ISSUE #82: Core: refactor "privileged" modules
Some modules in core like `raw` and `inline_content` are native to the compiler and need privileged access to the `core::Context` struct.  Currently we just have a bunch of if statements that check for these names right in the eval function. It would be a lot nicer if these are expressed in some more flexible way that also would allow us to add `ArgInfo` and other things needed for better potential IDE integration

(Note that this is not super important right now, especially if we want to refactor the eval function later anyway.

CMDJojo:
This PR aims to restructure the native packages, those which are written in rust in the source code to Modmark, to use the same interface as external packages loaded from wasm. This is done mainly by changing `wasm_module: Module` in the `Package` struct to an enum holding two options, either the wasm module or a flag telling the core that the package is implemented natively.

Some work has already been done, and now all packages are under the same interface. We no longer have lots of `if name == "raw" {...}; if name == "parse_inline" {...} ...`, the native packages are in the same list as the normal packages and are searched for in the same way. Some macros has also been added to simplify the declaration of such native packages.

There is still some work to be done:

- [x] Change `(from, NATIVE)` to something better (currently, native modules encode their transform to go from the target module to the output format `NATIVE`)
- [x] ~~Maybe move native modules to another list, since they don't have an output format, just a source format~~ They are in the same map but with two different cases for the value
- [x] ~~Maybe split the different `Element` types to wrap opaque structs, so pattern matching isn't needed as much~~ Since we are still unsure if we may support native parents, this is skipped for now. It may be changed later if we decide to only allow modules, but if not this may be unnecessary
- [x] ~~Come to a decision on how/if args are allowed for native Parent nodes~~ See above, skipped for now
- [x] General clean-up

Resolves #82

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -1,15 +1,19 @@
-use crate::Element;
-use parser::ModuleArguments;
-use serde::{Deserialize, Serialize};
 use std::{
     collections::HashMap,
     io::{Read, Write},
 };
+
+use serde::{Deserialize, Serialize};
 #[cfg(feature = "native")]
 use wasmer::Cranelift;
-use wasmer::{Instance, Store};
+use wasmer::{Instance, Module, Store};
 use wasmer_wasi::{Pipe, WasiState};
 
+use parser::ModuleArguments;
+
+use crate::package::PackageImplementation;
+use crate::std_packages::StandardPackages;
+use crate::Element;
 use crate::{ArgInfo, CoreError, OutputFormat, Package, PackageInfo, Transform};
 
 #[derive(Debug)]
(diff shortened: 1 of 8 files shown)


CMDJojo:
I think this is ready for review now, since the goals with the PR has been achieved. It may also be blocking #80 since that changes how packages are important, so it might be good to merge this before a PR is made for that.

core/src/std_packages.rs
```diff
@@ -0,0 +1,93 @@
...
+            }
+        ] => native_set_env,
+    };
+}
+
+// Here, all standard packages are declared. The macro expands to one function
+// which takes a &mut Context and loads it with the given standard package. It is important that the
+// package with a given name both is in a folder with that name, containing a cargo package with
+// that name. Otherwise, the module won't be found
+define_standard_package_loader! {
```
adelhult, line 34: Might be nice to move `define_standard_package_loader` above the native_package macro expressions since most of the time people will add non-native packages and being at the very top makes it a bit easier to find!

core/src/context.rs
```diff
@@ -281,78 +367,178 @@ impl Context {
...
+            ));
+        }
+        Ok(collected_args)
+    }
+}
+
+/// An enum containing one of two data types. Note that this is similar to `Result`, minus the
+/// convention that the right value is an error type.
+#[derive(Debug, Clone)]
+pub enum Either<L, R> {
```
adelhult, line 505: Perhaps we could have a special purpose type like `TransformOutput` instead since that's our only use case of `Either`. But I also like having a general either, so I'm not super convinced about that idea.  But, you might want to consider using the [either crate](https://crates.io/crates/either) instead.. Then we would have a bit less code to maintain.

adelhult (approved):
Looks good, left two small comments but I think the PR looks great!

core/src/std_packages.rs
CMDJojo: Fixed, I moved it up

core/src/context.rs
CMDJojo: I switched to use the Either crate. I thought it was the perfect fit for this use case, and that's why I opted for that instead of a custom type.

PR #84, opened 2023-02-22
GH-80: Fix memory leaks

### LINKED ISSUE #80: Rendered html output runs out of memory

adelhult:
This PR resolves #80. 
Comparing the memory footprint of the playground of this branch (left) to main (right) after 50 rerenders of an empty document you notice that the new version is still around 7-8mb while our current implementation has grown to 66mb.  
image

I got afraid that this issue would end up very hard to fix with proper web support without having to resort to recompiling all modules every time they are used. However, after a bit of digging in the `wasmer::js::module` I found that it seems like even though Wasmer requires the `Module::new` method to borrow a store it does not actually use it for anything. So we can create a dummy store when first compiling modules for web (and in the case of native platforms just use an Engine that we store in the context object) and then later create a new store each time we create a WebAssembly module instance.

I'm not entirely convinced that this is the way to go, and we will likely need to do a even bigger rewrite  
(and potentially store `Package`s in a JS context) to make chromium happy. But, we will deal with that issue another day.

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -205,9 +205,6 @@ impl Context {
                 body: _,
                 inline: _,
             } => {
-                // First, search for a package with transform to output format NATIVE
-                // If not found,
-
                 let Some((_, package)) =
                     self.transforms.get(name).and_then(|t|t.find_transform_to(output_format))
                      else {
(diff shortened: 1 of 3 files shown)


CMDJojo (approved):
Looks good! It will be a nicer experience now that it doesn't crash all the time :D hope this works

PR #86, opened 2023-02-22
Add error handling for modules

### LINKED ISSUE #85: Add error handling for modules
It would be nice if modules could report errors/warnings that would be collected throughout the evaluation cycle, which would possibly be outputted in the final document.

That the `stderr` for `wasm` modules is automatically collected, which turns it either to a warning or error depending on if `stdout` returns something valid, and possibly two std-modules `error` and `warning` that emits such errors/warnings.

CMDJojo:
This PR aims at adding error handling for modules. A module may fail for multiple reasons, and the default behaviour shouldn't be that no part of the document can be evaluated just because one module fails. It may not be obvious why a module failed, especially if one module invokes other modules etc.

This implementation of error handling works on this fundamental concept:
* The `Context` has a `CompilationState`, which keeps the state of the document compilation, including what errors and warnings have occurred

To facilitate error reporting, these native modules have been added:
* `[error]` which adds an error to the current context, and takes optional arguments `source` and `target`
* `[warning]` which adds an warning to the current context, and takes optional arguments `source` and `target`

Additionally, the `[error]` module will check if there is a package (usually the language package) handling `[__error]` modules, and will then generate such a module. This makes it possible to make the errors appear when they occur, at the place they occur, in the output document (for example, as red text describing the error), in addition to being collected. If no such `[__error]` transform exists, or if it itself is responsible for the error, no such module will be given.

Since wasi-modules have access to `stdout` and `stderr`, `stderr` was leveraged to allow for even easier warning/error reporting for developers:
* If a line is printed to `stderr`, it will become either a warning or an error (one warning/error per line)
* If `stdout` contains text, the lines in `stderr` are treated as warnings, and otherwise as errors. If `stdout` contains text (and the `stderr` lines thus are treated as warnings), the result will still be parsed as an `Element` and put into the tree. This is assuming that a warning occurred that didn't have to stop evaluation of the module (for example, an invalid language specified for a `code` module, but the text can still be displayed unformatted).

Additionally, calls to modules in wasmer results in a `Result`. If that fails, it is assumed that the module crashed (`panic`ed or similar), and its output won't be trusted to contain correct and uncorrupted data. Thus, it will generate an error and its output won't be used.

In addition to these things, this PR includes some basic error formatting for the HTML module as well as displaying the collected errors in the playground.

***Note:*** This PR is still a draft. It isn't finished yet, and several aspects and specifications described above are subject to change.

Todo before finished:
- [x] Determine if it is appropriate to save the input text to a module when an error occurred.
- [x] Change `(String, String)` (`(Source, Text)`) for a custom `Issue` type, possibly containing more data.
- [x] ~~Improving error display in HTML.~~ I think I'll leave that to the people working on the modules, but I have a placeholder implementation which works, and improvements can be made upon that.
- [x] Improving the `transform` function when it comes to error handling; there is some repeated code there at the moment, and a lot of boiler-plate that may be abstracted away.
- [x] Determine the necessity for the `target` field, and the necessity for adding additional fields.
- [x] General clean-up.

Resolves #85

### DIFF
--- a/modules/README.md
+++ b/modules/README.md
@@ -2,44 +2,59 @@
 
 Hey there!
 
-In this folder, you can find a bunch of useful packages that come with the compiler. But, what are packages, you ask? Well, they are programs that add support for transforming **module elements** (like `[math]` and `[table]`) as well as **parent elements** (like `**bold text**` and the document template) into a output format.
+In this folder, you can find a bunch of useful packages that come with the compiler. But, what are packages, you ask? Well, they are programs that add support for transforming
+**module elements** (like `[math]` and `[table]`) as well as **parent elements
+** (like `**bold text**` and the document template) into an output format.
 
-Packages work similarly to command-line programs: they receive a part of theto document in json format through stdin, modify it, and send the result back via stdout. Packages are loaded as `.wasm` files, which means they can be written in any programming language that supports WebAssembly (WASM) and the WebAssembly System Interface (WASI).
+Packages work similarly to command-line programs: they receive a part of the document in json format through stdin, modify it, and send the result back via stdout. Packages are loaded as `.wasm` files, which means they can be written in any programming language that supports WebAssembly (WASM) and the WebAssembly System Interface (WASI).
 
 For example, if you want to create your own package in Rust, compile it to `--target wasm32-wasi`.
 
-
 # Writing a package of your own
+
 If you wish to write your own package you need to handle two types of request that are specified as command-line arguments to your program.
 
 ## Manifest
-When your program is called with the arguments `$ ./my_program manifest` it expects to recieve a json object (sent via stdout) that explains how your package works and what transforms it supports. A manifest might look something like this:
+
+When your program is called with the arguments `$ ./my_program manifest` it expects to receive a json object (sent via stdout) that explains how your package works and what transforms it supports. A manifest might look something like this:
+
 ```json
 {
     "name": "Name of your package",
     "version": "0.1",
     "description": "A short description that explains the function of your package",
     "transforms": [
-            {
-                "from": "foo",
-                "to": ["html", "tex"],
-                "arguments": [
-                    {"name": "a_argument", "description": "This is the description for the argument 'a_argument'"},
-                    {"name": "another_argument", "default": "hello", "description": "..."}
-                ],
-            },
-            {
-                "from": "bar",
-                "arguments": [],
-            }
+        {
+            "from": "foo",
+            "to": [
+                "html",
+                "tex"
+            ],
+            "arguments": [
+                {
+                    "name": "a_argument",
+                    "description": "This is the description for the argument 'a_argument'"
+                },
+                {
+                    "name": "another_argument",
+                    "default": "hello",
+                    "description": "..."
+                }
+            ]
+        },
+        {
+            "from": "bar",
+            "arguments": []
+        }
     ]
 }
 ```
 
 ## Transforming an element
-When the program is called with `$ ./my_program transform <element_name> <output_format>` you will need to transform an element into the youdesired output format.
 
-You will be sent the element as json object via stdout and you respond by sending a json list of objects representing elements back.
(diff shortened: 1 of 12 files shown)


CMDJojo:
Updates for today:
* I made an `Issue` struct, holding the data for an issue (warning or error)
* I kept the `target` field after feedback from team member
* I added an `input` field, optionally holding the input data that caused an error
* There is a flag in `CompilationState` which tells the `transform` function if it should collect input or not
* I made a closure in `transform` which makes `Issues`, which simplifies the code a lot

modules/README.md
```diff
@@ -2,43 +2,82 @@
...
-You will be sent the element as json object via stdout and you respond by sending a json list of objects representing elements back.
+When the program is called with `$ ./my_program transform <element_name> <output_format>` you will need to transform an element into the desired output format.
+
+You will be sent the element as json object via stdout, and you respond by sending a json list of objects representing elements back.
+
+If you want to see some examples of what this json format looks like, you can check out the subdirectories or visit the online playground and select "JSON output".
+
+## Handling errors
+
+Errors occur in all pieces of code, including your packages. As a part of the sandbox we build to run your packages in, we have built in some error handling.
```
adelhult, line 63: Typo, "built". But might be better to skip the sentence about entirely since the error system is not really coupled to exclusively the playground.

modules/html/src/example.json
```diff
@@ -1,12 +1,5 @@
...
-        {
-            "name": "__text",
-            "data": "Hello, world",
-            "arguments": {},
-            "inline": true
-        }
-    ]
-}
\ No newline at end of file
+    "name": "__error",
```
adelhult, line 2: We might want to revert this change (or maybe just remove the `example.json` file)

modules/html/src/main.rs
```diff
@@ -1,10 +1,20 @@
-use serde_json::{from_str, json, Value};
 use std::{
     env,
     fmt::Write,
     io::{self, Read},
 };
 
+use serde_json::{from_str, json, Value};
+
+macro_rules! raw {
```
adelhult, line 9: Nice : )

modules/html/src/main.rs
```diff
@@ -93,32 +105,64 @@ fn transform_heading(heading: Value) -> String {
...
-            .replace('\n', "\\n");
-        format!(r#"[{{"name": "raw", "data": "{s}"}}]"#)
+        let s = escape(s);
+        format!("[{}]", raw!(s).to_string())
     } else {
         panic!("Malformed text module");
     }
 }
 
+fn escape(text: &str) -> String {
```
adelhult, line 153: maybe a char instead of a &str as parameter

adelhult (approved):
Looks good, great work! Left a small nit picky comments but nothing too important.

modules/html/src/example.json
CMDJojo: Yes, I will revert it

modules/html/src/main.rs
CMDJojo: Haha thanks! I have changed some other parts of the code using `raw` nodes as well, in addition to my own code

modules/html/src/main.rs
CMDJojo: Hmm, I think this is the most common use case, but I can do a separate function which just escapes a character, which this function delegates to. I doubt it will be very useful, and I think this may be more efficient since it is just one string and not multiple strings that has to be combined, but if you think there will be a need for a function taking a char, I will do it!

modules/README.md
CMDJojo: Hmm, I don't really understand what you are aiming for now. We do build a sandbox environment for the packages to run, right? It isn't specific to the playground, we use Wasmer in both native and web-targeted builds. With "sandbox", I am referring to the Wasmer runtime. Do you think we should skip the information that errors will be caught by our sandbox? Or is it better to rephrase it?

Issue #89, opened 2023-02-23
Add warnings from build script for failed module builds


Forsinge:
**Is your feature request related to a problem? Please describe.**
Currently the build script is quiet when modules fail to build. This makes it harder to troubleshoot.

**Describe the solution you'd like**
Warnings that specify which modules failed to build, using ``println!(cargo:warning=___)``

PR #90, opened 2023-02-23
GH-89: Add build script warnings for failed modules builds

Forsinge:
Resolves: #89 

This PR adds warnings in the build script for when modules fail to build.

### DIFF
--- a/core/build.rs
+++ b/core/build.rs
@@ -35,16 +35,19 @@ fn main() {
                 &out_path,
             )
         })
-        .for_each(|mut f| {
-            f.wait().expect("failed to launch wasm build");
+        .for_each(|(mut f, name)| {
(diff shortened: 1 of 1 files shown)


adelhult (approved):
Looks good!

CMDJojo:
Isn't it better if the compilation fails alltogether (i.e an warning instead of an error), if possible?

CMDJojo:
(If that isn't possible or hard to do, feel free to merge this; the compilation will fail anyways and thus the logs may show what module is problematic, which is much better than we had before)

Forsinge:
I think it may be possible, at the very least with a ``panic!``. One disadvantage of doing that is when several modules fail to build. In this case you would only be notified about one.

CMDJojo:
Yeah, that is true. Keep it as we have it now then. Good work!

PR #92, opened 2023-02-23
GH-91: Add support for disabling the bundling of standard packages

### LINKED ISSUE #91: Feature request: Make packages conditionally bundled
It would be nice to be able to distribute the program without the standard packages bundled. Since the standard packages currently are built-in and included in the binary by copy-pasting using `include_bytes!`. This has its pros and cons, it is a nice solution since the packages doesn't need to be installed separately from the binary itself, so it is easier to use for the user. It is also nice since they are "just there" in the web as well, no other files to find and download etc.

The con is that the binary gets much larger (almost 50% larger, with 2 modules, currently), and since we are only at the start of developing modules, it is to assume that this will be even more of a hit later on. If we bundle modules for every different output format we support, and different features such as tables, citation etc, lots of people may only use a small fraction of what we offer. It might be better, for them, to not have all modules pre-bundled but rather installed-on-demand with a package manager. Also, with the package manager being closer to completion than ever, I feel like this is appropriate.

Screenshot 2023-02-23 at 17 27 41

*modmark compiled both with and without bundled packages, under release configuration, with packages ~14.5MB, without packages ~9.8MB*

I think we can use features here. We may have a feature `bundle_std_packages` which is enabled by default, and signals that the standard packages should be bundled. This would need to be supported by both `cli` and `playground` which both import `core`, so that they supply the `bundle_std_packages` feature if not disabled. Additionally, if `bundle_std_packages` isn't enabled, the build script that builds the packages may be skipped to improve build times.

CMDJojo:
This PR adds support for disabling bundling of standard packages. There is now a feature `bundle_std_packages` which is enabled by default, which makes sure the packages are built and bundled. If you build either `cli` or `playground` with `--no-default-features`, it will be disabled and the packages will not be bundled.

Resolves #91

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -6,10 +6,13 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-parser = {version= "*", path = "../parser/"}
-core = {version= "*", path = "../core/"}
+parser = {version= "*", path = "../parser/" }
+core = {version= "*", path = "../core/", default-features = false, features = ["native"]}
 clap = { version = "4.1.4", features = ["derive"] }
 notify = "5.0.0"
 crossterm = "0.26.0"
 thiserror = "1.0.38"
 directories = "4.0.1"
+
+[features]
+default = ["core/bundle_std_packages"]
--- a/playground/web_bindings/Cargo.toml
+++ b/playground/web_bindings/Cargo.toml
@@ -11,13 +11,10 @@ crate-type = ["cdylib", "rlib"]
 [dependencies]
 parser = { version = "*", path = "../../parser/" }
 thiserror = "1.0.38"
-core = { version = "*", path = "../../core/", default-features = false, features = ["web"] }
+core = { version = "*", path = "../../core/", default-features = false, features = ["web", "bundle_std_packages"] }
 wasm-bindgen = "0.2.63"
 # The `console_error_panic_hook` crate provides better debugging of panics by
 # logging them with `console.error`. This is great for development, but requires
 # all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
 # code size when deploying.
 console_error_panic_hook = { version = "0.1.6" }
-
-[features]
-default = ["core/bundle_std_packages"]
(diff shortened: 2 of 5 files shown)


cli/Cargo.toml
```diff
@@ -6,10 +6,13 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-parser = {version= "*", path = "../parser/"}
-core = {version= "*", path = "../core/"}
+parser = {version= "*", path = "../parser/" }
+core = {version= "*", path = "../core/", default-features = false, features = ["native"]}
```
adelhult, line 10: Since the CLI is a executable and not a library I think it would be more conventional not to create a separate feature flag for this crate but instead just use the defaults.

playground/web_bindings/Cargo.toml
```diff
@@ -9,12 +9,15 @@ crate-type = ["cdylib", "rlib"]
...
 # The `console_error_panic_hook` crate provides better debugging of panics by
 # logging them with `console.error`. This is great for development, but requires
 # all the `std::fmt` and `std::panicking` infrastructure, so isn't great for
 # code size when deploying.
-console_error_panic_hook = { version = "0.1.6" }
\ No newline at end of file
+console_error_panic_hook = { version = "0.1.6" }
+
+[features]
+default = ["core/bundle_std_packages"]
```
adelhult, line 23: Same deal here, the bindings would not need their own features, they can just choose directly when they import core.

adelhult (changes requested):
Sounds like a really good idea to expose a feature flag for the core crate! But, I think it's a bit overkill to have separate features for the cli and web_bindings crate.

cli/Cargo.toml
CMDJojo: I strongly think that it is reasonable to opt out of bundled packages in the CLI. The executable size grows by 50% at the moment, and when adding more packages it will grow even more, and some people may just prefer to use the package manager to download the packages they need rather than having more than they'll ever use, taking up space in the binary and giving slower startup times. So I believe there is a reason to have two distributions, one with bundled packages and one without, so I am keeping it as-is and hope you will agree with me :)

adelhult (approved):

Issue #81, opened 2023-02-20
CLI: Infer output format based on the output file's extension


adelhult:
**Is your feature request related to a problem? Please describe.**
If you for example specify the output file to be "test.tex" it would be nice if the cli could infer the output format to be "tex" or "latex" instead of always using the "html" default.

PR #95, opened 2023-02-23
GH-81: Infer output format based on file extension

adelhult:
Resolves GH-81.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -11,8 +11,8 @@ use crossterm::{
 use error::CliError;
 use notify::{Config, Event, PollWatcher, RecommendedWatcher, RecursiveMode, Watcher, WatcherKind};
 use parser::{parse, Ast};
-use std::env;
 use std::io::{stdout, Write};
+use std::{env, path::PathBuf};
 use std::{fs, path::Path};
 use std::{fs::File, time::Duration};
 
@@ -20,18 +20,13 @@ use std::{fs::File, time::Duration};
 #[command(author, version, about)]
 struct Args {
     #[arg(index = 1, help = "Path to input file")]
-    input: String,
+    input: PathBuf,
 
     #[arg(index = 2, help = "Path to output file")]
-    output: String,
+    output: PathBuf,
 
-    #[arg(
-        short = 'f',
-        long = "format",
-        help = "The output format of the file",
-        default_value = "html"
-    )]
-    format: String,
+    #[arg(short = 'f', long = "format", help = "The output format of the file")]
+    format: Option<String>,
 
     #[arg(
         short = 'w',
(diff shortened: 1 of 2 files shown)


CMDJojo:
Should we infer output extension from format flag, if the output doesn't have an extension? Like should `modmark -f html in.mdm out` make a file `out.html`?

cli/src/main.rs
```diff
@@ -44,10 +39,30 @@ struct Args {
     dev: bool,
 }
 
+// Infer the output format based on the file extension of the output format
+fn infer_output_format(output: &Path) -> Option<OutputFormat> {
+    output
+        .extension()
+        .map(|ext| match ext.to_str() {
+            Some("tex") => Some(OutputFormat::new("latex")),
+            Some("html") => Some(OutputFormat::new("html")),
```
CMDJojo, line 48: Maybe add htm as well?

cli/src/main.rs
```diff
@@ -44,10 +39,30 @@ struct Args {
     dev: bool,
 }
 
+// Infer the output format based on the file extension of the output format
+fn infer_output_format(output: &Path) -> Option<OutputFormat> {
+    output
+        .extension()
+        .map(|ext| match ext.to_str() {
```
CMDJojo, line 46: `and_then` to not have to map

cli/src/main.rs
```diff
@@ -44,10 +39,30 @@ struct Args {
...
+}
+
 fn compile_file(args: &Args) -> Result<Ast, CliError> {
     let source = fs::read_to_string(&args.input)?;
     let mut ctx = Context::default();
-    let output = eval(&source, &mut ctx, &OutputFormat::new(&args.format))?;
+
+    let Some(format) = args.format
+        .as_ref()
+        .map(|s| OutputFormat::new(&s))
```
CMDJojo, line 60: Hmm, you do `as_ref` and then take a ref, seems to me that you can replace `&s` to `s`, and then u are fine with just giving the function handle directly

CMDJojo (changes requested):
Seems nice, good work! I found some code style tings I think should be fixed before merging, as well as my previous comment which would be nice to implement, if you agree that that is a good idea

cli/src/main.rs
adelhult: Nope, sadly it doesn't work since `OutputFormat::new` expects a `&str` and I have `&String`.

CMDJojo: It is the same thing with automatic coersion, and `.map(|s| OutputFormat::new(s))` compiles on my machine

CMDJojo (approved):
Looks good now! (For other's information, we talked in private and concluded that while document programs often automatically add file extensions, command line tools doesn't so often, so we skipped it here for now). Ready to merge!

Issue #53, opened 2023-02-15
CLI: keep the Context object alive when using the "--watch" flag


adelhult:
**Is your feature request related to a problem? Please describe.**
Right now, a new context object is created every time the document is recompiled when using the `--watch` flag in the cli. This means that all packages will need to be reloaded every time the source text is updated. 

**Describe the solution you'd like**
It would be great if we instead could keep the context alive in-between changes.

PR #96, opened 2023-02-23
GH-53: Keep context alive when using --watch flag

adelhult:
Resolves GH-53.


CMDJojo (approved):
Nice, good job! This will be a nice change! Just out of curiosity, if we later on have documents inside of documents etc, this will keep the same context for all the documents, right? In that case, it might be something we want to change later on when we add support for multiple documents, but I think we can merge this anyways!

PR #98, opened 2023-02-25
Add standard code package

Forsinge:
Resolves: #60 

This PR adds the standard code package, which is meant for highlighting and displaying code snippets.

TODO:
- [x] Revise indent functionality
- [x] Allow specifying language either by extension or by full name
- [x] Handle write results (might be most convenient to just unwrap)
- [ ] Look into optimizing initialization of syntax sets (might not be necessary after playground updates)
- [x] Look into renaming and reordering arguments to be more convenient
- [x] Clean up


CMDJojo:
Beware of merge commits, I think our branch protection rules doesn't allow merge commits in main and I think I happened to get one once by mistake, and it was hard to get rid of. Just a tip, avoid them if possible. But nice work with the code module! Looking forward to seeing it when it is done!

CMDJojo:
I think Eli solved my merge commit issues by somehow just rebasing from main and that just flattened everything iirc, but just a heads up!

Forsinge:
Thanks for the heads up! I think I managed to solve it now, and I even renamed commits to reflect modules being renamed to packages!

packages/code/src/example1.json
```diff
@@ -0,0 +1,8 @@
+{
```
CMDJojo, line 1: It would be nice if you could turn these `example`s to full-on `tests`, so they can be in the repo and have a real place and purpose rather than just laying in the source folder. See `packages/html/tests` for examples.

packages/code/src/example2.json
```diff
@@ -0,0 +1,10 @@
+{
```
CMDJojo, line 1: Also here, see comment on `example1.json`

packages/code/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+    }
+}
+
+fn transform_code(to: &String) {
+    macro_rules! get_arg {
+        ($input:expr, $arg:expr) => {
+            if let Value::String(val) = &$input["arguments"][$arg] {
+                val
+            } else {
+                panic!("No theme argument was provided");
```
CMDJojo, line 63: I think this argument message is wrong, since you don't always get the "theme" argument using this macro, right?

packages/code/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+    style.push('\"');
+    style
+}
+
+fn get_highlighted(code: &str, lang: &String, tm: &String) -> (String, Option<Color>) {
+    let ss = SyntaxSet::load_defaults_newlines();
+    let ts = ThemeSet::load_defaults();
+    let syntax = match ss.find_syntax_by_token(lang) {
+        Some(sr) => sr,
+        _ => ss.find_syntax_by_token("py").unwrap()
```
CMDJojo, line 122: This is kind of strange, it seems like if tries to find the language from the user input but if it doesn't find it, it defaults to python. This means that if you do `[code cpp]` and `cpp` isn't found (if it is called `c++` instead), `python` formatting will be used. I don't think this is good; if the user specifically asks for a language syntax highligter and it doesn't exist, it should error. Maybe you could add `py` as the default argument instead?

CMDJojo: Also, it would be nice to know what languages are available, or how to find a list of them or something of the like

packages/code/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+        "github" => &ts.themes["InspiredGitHub"],
+        "solar_dark" => &ts.themes["Solarized (dark)"],
+        "solar_light" => &ts.themes["Solarized (light)"],
+        _ => &ts.themes["InspiredGitHub"]
+    };
+    let mut h = HighlightLines::new(syntax, theme);
+    let incl_bg = IncludeBackground::No;
+    let mut html: Vec<String> = vec![];
+
+    // avoiding lines() here because we want to include the final newline
```
CMDJojo, line 138: I am unsure what you mean by this, it seems to me that `lines()` does the same thing as you do, and I don't really know what you mean by including the final newline. When I tested both ways, it looks like I get the same result regardless if I do it this way or using `lines()`

CMDJojo: Oh well, I see that if you do it your way, you add a new `<br>` at the end. But I fail to see why that is needed since the result looks the same anyways, and you can always just append a `<br>` yourself... But maybe there is a good reason, please tell me, I am curious!

Even doing this:
```
[block_content]"
[code python]
def foo():
    1 + 1"
Text afterwards
```

which would make sure text comes immediately after the code block without an empty line in between, using `.lines()`, seems to work fine for me.

packages/code/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+        {
+        "name": "Standard code package",
+        "version": "0.1",
+        "description": "This package provides syntax highlighting in [code] modules",
+        "transforms": [
+            {
+                "from": "code",
+                "to": ["html"],
+                "arguments": [
+                    {"name": "lang", "description": "The language to be highlighted"},
```
CMDJojo, line 34: Hmm maybe have some default language/default to unformatted like \`\`\`-blocks? (Unformatted may be called `text` or something)

packages/code/src/main.rs
```diff
@@ -0,0 +1,144 @@
...
+    let mut h = HighlightLines::new(syntax, theme);
+    let incl_bg = IncludeBackground::No;
+    let mut html: Vec<String> = vec![];
+
+    // avoiding lines() here because we want to include the final newline
+    for line in code.split("\n").map(|s| s.trim_end_matches("\r")) {
+        let regions = h.highlight_line(line, &ss).unwrap();
+        html.push(styled_line_to_highlighted_html(&regions[..], incl_bg).unwrap())
+    }
+    return (html.join("<br>"), theme.settings.background);
```
CMDJojo, line 143: Code style thing: implicit return. Tip: Do `cargo clippy`

CMDJojo (changes requested):
Nice work @Forsinge ! Nice to get some color to our documents!

When testing around, I found some problems. Doing this:
```
[code b]
"abc"\def
ghi \ jkl
```
strangely removed "def" and "jkl" entirely from the output, while doing this:
```
[code c]
"abc"\def
ghi \ jkl
```
worked as expected. Additionally, I left some comments in the code with small improvements I think would make the package/module even better! (Also, it might be nice with a small guide/smth on what options are allowed, it is kind of hard to know what color schemes there are etc). But nice work, it was fun to play around with! Rebasing with main would make the testing experience more fluent, I think.

(Also, the compilation of the document takes ~250ms, but yeah, I know you've noticed that)

packages/code/src/main.rs
Forsinge: It gives different results when using delimiters.
For example:
```
[code python]{

def foo():
    print("foo")

}
```
When using ``lines()`` the empty row at the end will not be included, while it is included in the current implementation.

CMDJojo: Oh okay. Clever solution! Keep it as it is then :)

packages/code/src/main.rs
Forsinge: Fair point. The reason I did this was to avoid repeated panics as the user is typing out the argument. If this isn't something we should care about, I can definitely change it. I agree that silently defaulting to python like this can cause unwanted behaviour though.

As for available languages, I haven't found a list yet. I know it uses Sublime Text's default syntax definitions so maybe I can find something there. If I do find something, where would I put this information?

Forsinge: Is there maybe a good way to use your system for warnings for this?

CMDJojo: You could also do warnings, it is your choice, and I guess that you'll have to default to outputting text without syntax highlighting then. Warnings sound good as well! You do them by printing your warning text using `eprintln!(...)` and then continuing as normal, printing your JSON output to `stdout`

CMDJojo (approved):
Looks good! Let's go! Good work. You are officially the first one who is using warnings afaik, nice!

Issue #45, opened 2023-02-13
Playground cannot load packages in Chromium-based browsers


axoxelol1:
**Describe the bug**
Loading packages does not work in chromium.

**To Reproduce**
1. Build and view playground in a Chromium-based browser
2. Choose HTML output or click on view packages
3. See error in console

This is currently working in firefox and webkit and should also work in chrome. Maybe not the highest priority for now but needs to be investigated.

**Screenshot**
![image](https://user-images.githubusercontent.com/40837400/218471455-e9cb7e35-6d04-4e03-9c32-222e330f631c.png)

CMDJojo:
Do you think this may be solved by wrapping all calls to loading wasm and calling wasm-bindings in `async`/`await` in Javascript on the playground, or do you think it needs to be implemented by doing async rust? I'm thinking that wrapping everything in `async`/`await` will make it run in another JS thread, and that might be enough to stop these errors. (Of course, it is nice to do things like loading packages in parallel, but this might be a way to make it able to load the main bindgen files)

adelhult:
Related to our issue, will take a closer look once I have time: https://github.com/wasmerio/wasmer-js/issues/314

adelhult:
Related to our issue, will take a closer look once I have time: https://github.com/wasmerio/wasmer-js/issues/314

adelhult:
Bug hunting update 🐛:

The reason we can't use the playground in Chromium based browsers is that when using `wasmer::Module::new` to compile WASM modules that method in itself calls the **blocking** JS `WebAssembly.Module` constructor.  And Chromium would very much prefer us to use the asynchronous JS function `WebAssembly.compile` instead.

I've played around with bypassing wasmer and using `js_sys` to call the asynchronous function instead. That could would look something like this:

```rs
async fn compile(data: &[u8]) -> js_sys::WebAssembly::Module {
	let js_bytes = unsafe { js_sys::Uint8Array::view(data) };
	let promise = js_sys::WebAssembly::compile(&js_bytes.into());
	let result = wasm_bindgen_futures::JsFuture::from(promise).await;
	result
		.unwrap()
		.dyn_into::<js_sys::WebAssembly::Module>()
		.unwrap()
}
```

And as seen in [wasmer-js](https://github.com/wasmerio/wasmer-js/blob/80b158341965d6d3203f68ae2be26d5b6858d65b/src/wasi.rs#L164) it seems like we then can convert the WebAssembly::Module into a wasmer::Module.

[...]

PR #99, opened 2023-02-25
Improve playground

### LINKED ISSUE #97: Improve playground
The playground should be improved by fixing a bunch of bugs related to the different output formats and the packages tab should not just debug print the context object but instead have a proper interface.

adelhult:
General improvements to the playground to make it a bit more demo worthy. Resolves #97.
* [x] Add package inspection tab, see screenshot 
* [x] Replace `<textarea>` element with the [Ace text editor](https://ace.c9.io/), see screenshot 
* [x] Fix broken mobile layout
* [x] Fix animation when sliding in between views
* [x] Add iframe render preview option (this allows us to preview documents templates that contains imports and styling.
* [x] Use 'Inter' font for menus and rendering 
* [x] Update editor only after a set amount of time to avoid to many rerenders 
* [x] Show module errors in playground
* [x] Show module warnings in playground
* [x] Add a "other output formats" option
* [ ] ~~Remove the Document element when previewing without a iframe~~ I'll fix this later.

_New package inspection view_
image

_New text editor_
image


CMDJojo:
Nice work with the playground! It is nice to update the look a bit, and I think your work is great! The packages view is amazing!

I know this is under development, but I've tested it out on mobile a little bit and noticed some things that I wanted to highlight for you:
 - The editor provides a button with three dots by the line the cursor is in, providing some ugly command-palette, where some of the buttons doesn't work while some do. This is strange, and it would be amazing if it somehow just could use the device default/built-in behaviour when it comes to selecing, copying etc. image
 - The editor doesn't support moving around the cursor with the touchpad-like interface which appears when holding the space bar. This makes editing extremely painful.
 - The "compiled in" is transparent. image
 - It would be nice if the document updated all the time, async, not just after a grace period. I think it would be possible to make it async and have two (atomic) counters. Assume you want to update the view, you could get-and-inc one counter and get a sequence number for that task. When a task finishes, it can compare-and-set-on-gt to see if a more recent view is already rendered, and if not, render it. That would put the tasks in a background thread, not slowing down the UI, and hopefully make it feel more fluent.

Also, have you tested calling all web-binding-functions wrapped in async/await, and seen if that solves the Chrome issue we had (#45, I think I commented it on that issue as well)? I am thinking that if all the wasm code (rust code, linking to our core and context and wasmer etc) run in async, its subtasks (like calling WebAssembly.compile) may still be blocking but not block the main thread.

Anyways, good job! These were just some small thoughts and feedback that arose when playing around in the preview

CMDJojo:
> calling all web-binding-functions wrapped in async/await

I tested this out with one simple test, just now, and it doesn't seem to work. Good news is that Wasm has support for [multiple threads](https://web.dev/wasm-threads/), so if we don't do another workaround (like using JavaScript promises, which wasm-bindgen actually does [generate binds](https://rustwasm.github.io/wasm-bindgen/reference/js-promises-and-rust-futures.html) for, or writing our own small JS library).

But anyways, as said, good job! (Now that I am on desktop, I notice that I get a click-cursor when I hover the "Compiled in 3ms", is that supposed to happen? Does anything happen when I click it?)

playground/static/main.js
```diff
@@ -0,0 +1,288 @@
...
+const editorOptions = {
+    fontFamily: "IBM Plex Mono",
+    fontSize: "12pt"
+};
+
+let editor = ace.edit("editor");
+editor.setOptions(editorOptions);
+editor.session.setUseWrapMode(true);
+
+//  EWarnings and errors
```
CMDJojo, line 16: Typo

CMDJojo (approved):
Great work! Even though the new playground is worse on mobile than the previous playground, I think it overall is an improvement.

Issue #93, opened 2023-02-23
Allow modules in headings


Forsinge:
**Is your feature request related to a problem? Please describe.**
Headings can only contain text and tags at the moment. This should be extended to also allow modules.

**Describe the solution you'd like**
A simple way to solve it is to incorporate ``parse_paragraph_elements`` in ``parse_heading`` . This makes headings support text, smart punctuation, tags and inline modules.

PR #100, opened 2023-02-25
GH-93: Allow modules in headings

Forsinge:
Resolves: #93 

This PR extends the functionality of headings to allow modules. This is done by using ``parse_paragraph_elements`` inside ``parse_heading``.

Example: ``# This heading has a cool equation: [math]x^2=4``

### DIFF
--- a/parser/src/lib.rs
+++ b/parser/src/lib.rs
@@ -204,29 +204,18 @@ fn parse_document_blocks(input: &str) -> IResult<&str, Vec<Ast>> {
 ///
 /// returns: The heading node, if a successful parse occurs, otherwise the parse error
 fn parse_heading(input: &str) -> IResult<&str, Heading> {
-    let res = map(
+    map(
         pair(
             verify(take_while1(|c| c == '#'), |s: &str| {
                 s.len() <= u8::MAX as usize
             }),
-            preceded(space0, parse_heading_text),
+            preceded(space0, parse_heading_text).and_then(parse_paragraph_elements),
         ),
-        |(start, body)| (start, body),
-    )(input);
-
-    return match res {
-        Ok((str, (start, body))) => {
-            let (_, elements) = parse_paragraph_elements(body).unwrap();
-            Ok((
-                str,
-                Heading {
-                    level: start.len() as u8,
-                    elements,
-                },
-            ))
-        }
-        Err(e) => Err(e),
-    };
+        |(start, elements)| Heading {
+            level: start.len() as u8,
+            elements,
+        },
+    )(input)
 }
 
 /// Parses the text for a heading, consuming until a line ending is found.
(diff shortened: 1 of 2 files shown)


Forsinge:
This assumes ``parse_paragraph_elements`` will never produce an error if given a body by the heading parser. Is there any case where this would fail?

parser/src/lib.rs
```diff
@@ -204,22 +204,29 @@ fn parse_document_blocks(input: &str) -> IResult<&str, Vec<Ast>> {
...
-    map(
+    let res = map(
         pair(
             verify(take_while1(|c| c == '#'), |s: &str| {
                 s.len() <= u8::MAX as usize
             }),
             preceded(space0, parse_heading_text),
         ),
-        |(start, body)| {
-            let mut elements = tag::extract_tags(vec![Text(body.into())]);
```
CMDJojo, line 215: Isn't it enough just to replace the `tag::extract_tags` and the `smart_punctuate` with `parse_paragraph_elements`, and that's everything we need to do? Right? (And change map to map_parser)

CMDJojo (changes requested):
Nice work, it is good to have modules correctly parsed within headings, but I do think it may be even easier to do in another way

CMDJojo:
> This assumes `parse_paragraph_elements` will never produce an error if given a body by the heading parser. Is there any case where this would fail?

I think it currently won't fail but I also think that we shouldn't assume that it won't. If we change the code later on or so, I think it is best to not let the entire parser crash on some input (or, if it crashes on some obscure input today, to have it crash randomly on an user that happens to enter that input). I think nothing else of our code force-unwraps, and it would be very nice if this code can be re-written without it.

parser/src/lib.rs
Forsinge: Good idea! I didn't know ``map_parser`` worked like that. Now that I'm reading about it, it seems like it may even be possible to just use ``.and_then(parse_paragraph_elements)`` on top of ``preceded``.

CMDJojo:
Nice, it looks much better now! Good job!

CMDJojo (approved):

CMDJojo:
I think it is ready to merge!

Forsinge:
A question before I actually merge this. Since I used ``space0`` it currently isn't possible to output a heading with spaces in the start. Is this something we should try to change?

CMDJojo:
> A question before I actually merge this. Since I used `space0` it currently isn't possible to output a heading with spaces in the start. Is this something we should try to change?

I don't think so. The normal behaviour of markdown is that `#heading` is equivalent with `# heading` and to me that makes sense. I don't see any reason to have a heading with spaces in the start. If you want to allow that, you may say that you remove up to 1 space but never more, so that two spaces becomes one space in the heading.

CMDJojo:
I tested it just now and HTML also removes spaces at the start of headings, so if we actually capture it, it will probably get removed by the output format anyways.

image

Forsinge:
Alright. We could look into this later on then if we find a reason to change it.

Issue #78, opened 2023-02-20
Rename the "module" directory into "packages" 


adelhult:
Modules have been split into two different concepts , "modules" and "packages". The name of the folder containing the built-in packages should reflect this change.

PR #104, opened 2023-02-27
GH-78: Rename modules into packages

CMDJojo:
This PR aims to rename `modules` into `packages`. Previously, `modules` referred to both the wasm-bundles and the things you invoked from text like `[this]`. Now, `packages` is aimed at referring solely to the wasm packages/bundles/files containing, among other things, module transformations.


adelhult (approved):

CMDJojo:
We might have missed some parts, and the terms in the documentation might be mixed up in some places, but I think we have to look through and re-write parts of the documentation anyways since things are outdated, and we have changed terminology, syntax and semantics a bit after writing some parts.

Issue #106, opened 2023-02-28
Fix various bugs in parser


CMDJojo:
**Describe the bug**
Some small problems in the parser when parsing corner cases has arisen from extended use:
 * When re-parsing the empty string as inline content, it fails:
   `[inline_content]""` → `Parsing error: NomError( "Error { input: \"\", code: Many1 }", )`
 * When parsing multiline modules where the first row is blank, instead of having no body, it consumes the entire document:
   ```
   [module]

   this should
   be outside

   of the module
   ```

**To Reproduce**
See minimal examples above

PR #107, opened 2023-02-28
GH-106: Fix various bugs in parser

CMDJojo:
Resolves GH-106 and the problems described in there, as well as adding some tests that verifies that the fixes work correctly, and some small general improvements in the parser.


hugomardbrink (approved):
LGTM!

PR #108, opened 2023-02-28
Add basic latex support

bruhng:
Resolves GH-58:
Adds basic latex support

Images showing all the modules working below:

image


packages/latex/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+        from_str(&buffer).unwrap()
+    };
+
+    match from {
+        "__bold" => transform_tag(input, "textbf"),
+        "__italic" => transform_tag(input, "emph"),
+        "__superscript" => transform_tag(input, "textsuperscript"),
+        "__subscript" => transform_tag(input, "textsubscript"),
+        "__underlined" => transform_tag(input, "underline"),
+        "__strikethrough" => transform_tag(input, "sout"), //fixme: needs a package to use
```
CMDJojo, line 46: You could (if you want) add this to your prelude when wrapping the `__document`, but if you rather want to wait until we get env done, do so!

packages/latex/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+        "__verbatim" => transform_block(input, "verbatim"),
```
CMDJojo, line 47: I think this should use the inline `\verb|...|` since this is the verbatim tag, which (most likely) doesn't span multiple lines and cannot span multiple paragraphs

packages/latex/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+}
+
+fn transform_heading(heading: Value) -> String {
+    let mut result = String::new();
+    result.push('[');
+
+    let Value::String(s) = &heading["arguments"]["level"] else {
+        panic!();
+    };
+    let level = s.parse::<u8>().unwrap().clamp(1, 6);
```
CMDJojo, line 95: I think LaTeX (by default) only has `section`, `subsection` and `subsubsection`. Clamping your level to `6` will be able to generate `subsubsubsubsubsection`, which doesn't exist. It is possible to create more of such commands, so you will either need to include such commands (complicated, maybe bloated and maybe not so useful, especially without env variables) or clamp to 3.

packages/latex/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+    let mut subs = String::new();
+    if level > 1 {
+        subs.push_str(&"sub".repeat((level - 1) as usize));
+    }
+    
+
+    write!(result, r#"{{"name": "raw", "data": "\n\\{subs}section{{"}},"#,).unwrap();
```
CMDJojo, line 102: Creative work with `{subs}section` :D love it

packages/latex/src/main.rs
```diff
@@ -0,0 +1,222 @@
...
+    match from {
+        "__bold" => transform_tag(input, "textbf"),
+        "__italic" => transform_tag(input, "emph"),
+        "__superscript" => transform_tag(input, "textsuperscript"),
+        "__subscript" => transform_tag(input, "textsubscript"),
+        "__underlined" => transform_tag(input, "underline"),
+        "__strikethrough" => transform_tag(input, "sout"), //fixme: needs a package to use
+        "__verbatim" => transform_block(input, "verbatim"),
+        "__paragraph" => transform_paragraph(input),
+        "__document" => transform_block(input, "document"),
```
CMDJojo, line 49: You will probably need something more in the prelude to make the document compile,,,

CMDJojo (changes requested):
Nice job with the LaTeX package! I see many creative and good solutions here, so good job! Sadly, the document produced doesn't compile. There is no prelude, and I would suggest adding `\documentclass{article}` as a start at least. Try to take the example ModMark document pre-loaded in the preview, remove the `table` module (which is not yet available for latex) and try to compile the document to LaTeX. Then, copy-paste it into Overleaf or any editor of your choice and see if it works. For me, it doesn't... It doesn't have to look perfect, but it would be nice if it was at least readable.

Screenshot 2023-02-28 at 17 34 25

But good job, it is almost done!

bruhng:
image

image

It now compiles and uses the correct version of verbatim.

CMDJojo (approved):
Looks good! Great!

One small note is that < becomes upside down exclamation mark (it seems like, I had no clue) and > becomes upside down question mark. As well as that it might be worth giving a warning if someone uses a heading level higher than 3 (which is the maximum supported, you can easily give a warning by just `eprintln!()`-ing it. But you don't have to fix that. Your code looks good and it generates documents with complicated stacked tags (I saw your ulem import, great!), so I will approve this! 

Good work!

PR #109, opened 2023-02-28
GH-97: Clean up playground JS and add example document

CMDJojo:
This PR aims to clean up the JavaScript code in the playground a bit, increasing performance by disabling type coercion and disabling `console.log`ging things every compile. The PR also adds an example document, and when the playground is loaded, that document is loaded and the editor is changed into "Render HTML" mode.

Resolves GH-97


adelhult (approved):
Looks great!

Issue #116, opened 2023-02-28
Math module


CMDJojo:
**Is your feature request related to a problem? Please describe.**
It would be nice with a `[math]` module which has support for multi-line and inline math

**Describe the solution you'd like**
For HTML output, MathML would be appropriate. LaTeX syntax is standard for math typesetting, so it would be preferable if we convert LaTeX to MathML for insertion into HTML. For LaTeX output, I think we can just keep the input as-is, and wrap it in a `\begin{equation}`-`\end{equation}`-block

**Describe alternatives you've considered**
We could use a web library that actually renders LaTeX, but I think MathML is better since it is more native to browsers.

**Additional context**
The [latex2mathml](https://crates.io/crates/latex2mathml) crate seems like a good fit for conversion to MathML in Rust.

PR #117, opened 2023-03-01
GH-116: Add math module

CMDJojo:
This PR implements a `[math]` module, usable both inline and multi-line, with both `HTML` and `LaTeX` as output formats.

For the HTML output format, [mathjax](https://www.mathjax.org/) is imported by adding a `<script async src=...>` tag just before the math content. This can be disabled by giving `false` as the `import` argument, such as `[math import=false]$x^2$`. mathjax allows displaying of MathML in browsers which lacks native support for that. [latex2mathml](https://crates.io/crates/latex2mathml) is used to convert LaTeX to MathML when the output format is HTML, and since that natively supports both inline-style and block-style math, the `inline` parameter is carried over.

For the LaTeX output format, `$...$` is used for inline math, making sure to escape any `$` in the input by replacing it with `\$`. For multiline math, `\begin{equation}....\end{equation}` is used.

The PR also includes tests with high branch coverage.

Resolves GH-116.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -6,9 +6,7 @@ members = [
     "package-tests"
 ]
 
-
-exclude = ["packages/table", "packages/html", "packages/latex", "packages/code", "playground/web_bindings"]
-
+exclude = ["packages/table", "packages/html", "packages/latex", "packages/code", "packages/math", "playground/web_bindings"]
 
 # We need to specifiy that we use the new resolver to
 # get wasmer to build correctly when using a workspace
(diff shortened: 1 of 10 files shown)


CMDJojo:
Note that entering invalid LaTeX, it will sometimes give an error that propagates from the `latex2mathml` library, as if you do a `\begin{align}` without an `\end{align}`, but in many simple cases, like `[math]x^`, `latex2mathml` will itself insert an `[PARSE ERROR: Undefined("EOF")]`, and since that isn't propagated to outside the `latex2mathml`, it is hard to actually propagate that to an actual `[error]`/`[warning]`, so I think we are better of letting that be as-is.

CMDJojo:
A little less than 2 months ago, Chromium got [mathml support](https://mathml.igalia.com/news/2023/01/10/mathml-in-chromium-project-is-completed/), so I am thinking of removing the import altogether (or, at least, default it not to import). What do you think?

CMDJojo:
I decided to disable importing Mathjax by default, see my previous comment. I think compatibility is high enough, but the option is still there if need be. Ready for review!

bruhng (approved):
Good work! Smart ideas to use mathml

PR #121, opened 2023-03-01
GH-120: Add -webkit-appearance: none;

### LINKED ISSUE #120: Add -webkit-appearance: none;
Currently, we allow the standard webkit appearance for the `option` tag. This makes it render strangely on Safari in comparison to on, let's say, Firefox. Here is a preview of how it currently looks:
image

Adding `-webkit-appearance: none;` to the CSS for the `option` tag makes it look more in line with the other browsers. It has a slightly different height due to no material icon being present, but I think it is okay anyways.
Screenshot 2023-03-01 at 13 26 52

We could try adding a material icon to get the same height or try fixing the height, but the first one isn't that easy (I've looked into it a little bit) and the second one is not a good solution. I think adding the simple `-webkit-appearance: none;` is a simple fix to the problem which will make the playground look better for now. If we are going to rewrite it in, let's say, React, everything will change anyways so I don't think we need to spend that much time on styling now.

CMDJojo:
This PR adds `-webkit-appearance: none;` to the `<option>` in the playground, which makes it look better on Safari on Mac. I think this is a simple improvement that is worth merging. See #120 for screenshots and more details. It also reformats the CSS file and removes a duplicate key `height` under the `#editor, #preview` selector. The duplicate key doesn't do anything and can thus be removed.

Resolves #120

### DIFF
--- a/playground/static/expand_more.svg
+++ b/playground/static/expand_more.svg
@@ -1,0 +1,1 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="48" viewBox="0 96 960 960" width="48"><path d="M480 711 240 471l43-43 197 198 197-197 43 43-240 239Z"/></svg>
--- a/playground/static/style.css
+++ b/playground/static/style.css
@@ -44,8 +44,8 @@ code {
 }
 
 button,
-select,
-input {
+input,
+select {
     display: flex;
     align-items: center;
     gap: 0.2rem;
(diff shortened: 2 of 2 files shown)


bruhng (changes requested):
I think we should still have any icon so it is clear that you can change it and that it is a list, with your change it just looks lite a block

CMDJojo:
> I think we should still have any icon so it is clear that you can change it and that it is a list, with your change it just looks lite a block

I do agree with you. However, the `select` can't contain material icons, so we would need to wrap the thing into another element containing the `select` and the icon. I have tried like this:
```html
            <a href="javascript:document.getElementById('selector').click()">
                <select name="selector" id="selector">
                    <option value="ast">Abstract syntax tree</option>
                    <option value="ast-debug">Debug AST</option>
                    ...
                </select>
                <span class="material-symbols-outlined">
                    expand_more
                </span>
            </a>
```
but that fails to open up the select when clicked, which means that you have to click on just the text itself to open up a select box. Additionally, it would look something like this:

Screenshot 2023-03-01 at 13 53 24

which I just think is ugly. I unfortunately see no way of removing all styling and then add an icon, while still being within the select element itself. I will continue looking at it, though...

CMDJojo:
image

The best I could achieve was this, which is adding a custom background image which is kind of a hack, but it works. It is still the material icon, rendered as svg, put as a background with additional padding added so it doesn't collide with the text. Is this satisfactory?

CMDJojo:
Edit: Made the icon a little bit smaller
image

bruhng (approved):
Looks better and has better ux! Nice!

CMDJojo:
> Looks better and has better ux! Nice!

Thanks for the push to make me find a better solution!

PR #123, opened 2023-03-01
GH-122: Escape table module rather than passing it as text

### LINKED ISSUE #122: Escape table module rather than passing it as text in the example document
In the example document, we use `[table]` in text form. Currently, we achieve this by doing `[__text][[table]]`

It is probably better to escape the `[table]` module with `\[table]`

We could keep it as-is, but that may be more confusing to new users which aren't comfortable with the language.

CMDJojo:
This changes the example document to use `\[table]` instead of `[__text][[table]]`, which is probably less confusing for new users.

Resolves #122


bruhng (approved):
Looks good

Issue #94, opened 2023-02-23
Table package improvements


axoxelol1:
There are multiple improvements that can be made for the table package. Quotes and backslashes cannot be used, this is because raw strings are currently being printed instead of serializing json. This will be an issue for many packages so discussing a good solution for this would be a good idea. The best way might be to construct the json with a serde_json macro then converting it to a string.

- [ ] Fix quote and backslash support
- [ ] Arguments for alignment
- [ ] Arguments for borders
- [ ] Option for first row to be headers

CMDJojo:
I would add "Strip whitespace" as a task as well, so that
```
[table]
a | b | c
```
is equivalent with
```
[table]
a|b|c
```

PR #124, opened 2023-03-01
GH-112: Add latex support to table

bruhng:
- Resolves: GH-112
Adds support for table->Latex

image

image

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -130,7 +130,7 @@ mod tests {
             description: "This package supports [table] modules".to_string(),
             transforms: vec![Transform {
                 from: "table".to_string(),
-                to: vec![OutputFormat::new("html")],
+                to: vec![OutputFormat::new("html"), OutputFormat::new("latex")],
                 arguments: vec![ArgInfo {
                     name: "col_delimiter".to_string(),
                     default: Some("|".to_string()),
--- a/packages/table/src/main.rs
+++ b/packages/table/src/main.rs
@@ -36,8 +36,8 @@ fn manifest() {
     .unwrap());
 }
 
-fn transform(from: &String, to: &String) {
-    match from.as_str() {
+fn transform(from: &str, to: &str) {
+    match from {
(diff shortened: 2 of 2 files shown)


CMDJojo (changes requested):
It is nice to have LaTeX output support for tables! However, I think there are some parts here that can be improved.

The task of the Table module is two-split:
 - Parse the content, pick out what the cells are, their content etc etc.
 - Render the content in the output format of choice

The first part, parsing, is probably the more complicated part. Rendering in HTML is just adding some <table>-tags and in LaTeX, it is adding some \begin{tabular} etc. The parsing part should be common between the two output formats; the syntax and semantics of the input to the table module should result in the same content, regardless of the output format. We currently have a pretty simple parser, but #94 aims to add a more complicated parser. I think it is a bad idea to just copy-paste the parser between the LaTeX and HTML implementation as you do now, since that could result in different results depending on the output format, which is not wanted.

You could do this by returning a custom node that contains the parsed content, which gets handled by this package later, like a `__table_ir` node, and define a transform `__table_ir` to HTML and LaTeX. Another alternative would be to just create a custom struct holding the parsed table, make a function that parses the table into that struct, and use that function both when outputting LaTeX and HTML. Both these ways would ensure that there is one common parser and that the parsing code is synchronized between the two different output formats.

bruhng:
https://github.com/modmark-org/modmark/pull/124#pullrequestreview-1320096925

Of course it would be better to link the parsing of the two, this was just a temporary solution to be able to render the example document in Latex.

bruhng:
Moved out the parsing part, is this better?

bruhng:
Fixed according to clippy

CMDJojo:
Yes, that seems much better! I would also love some tests to verify that it works correctly and as expected, so that if further changes are done, we can be sure that it didn't break anything or created invalid JSON.

CMDJojo:
I think we can close this since #150 is merged

Issue #52, opened 2023-02-15
Rename "core" to avoid conflicts with Rust core.


adelhult:
**Describe the bug**
We seem to get some conflicts with rust core due to our crate having the same name. I would suggest that we rename core
into "modmark_core".

CMDJojo:
Is this still relevant/something you think we should do? (I.e. is this still problematic, and if so, in what cases?)

Issue #48, opened 2023-02-13
Make the package manager dynamically download packages on compilation


hugomardbrink:
**Is your feature request related to a problem? Please describe.**
Download packages that are imported in the language

**Describe the solution you'd like**
Download .wasm files to the cache through the web that are requested from the package manager

**Additional context**
To make this work there need to be an consensus on where and how the URLs are handled, either:
- Core requests packages from the CLI and gets provided 
- The CLI gets the urls directly from the parser and downloads them before running core

To make this safe there should also be some safety-checks in place, this could be a JSON file with trusted packages that people can change with PRs. This would enable shorthands in the imports while pointing to the right .wasm file, although this would complicate development of packages.

There should also be question to the user while using the CLI if it wants to download a package.

We could also limit the imports to only github etc but this would also limit the potential of the language.

PR #126, opened 2023-03-01
GH-48: Add package managment via registry

hugomardbrink:
- Resolves: GH-48
- Resolves: GH-52

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,5 +1,5 @@
-use std::io::{stdout, Write};
-use std::{env, fs, fs::File, path::Path, path::PathBuf, sync::Mutex, time::Duration};
+mod error;
+mod package;
 
 use clap::Parser;
 use crossterm::{
@@ -7,16 +7,15 @@ use crossterm::{
     style::{self, Stylize},
     terminal, ExecutableCommand,
 };
+use error::CliError;
+use modmark_core::{context::CompilationState, OutputFormat};
+use modmark_core::{eval, Context};
 use notify::{Config, Event, PollWatcher, RecommendedWatcher, RecursiveMode, Watcher, WatcherKind};
 use once_cell::sync::Lazy;
+use std::io::{stdout, Write};
+use std::{env, fs, fs::File, path::Path, path::PathBuf, sync::Mutex, time::Duration};
 
-use core::context::CompilationState;
-use core::{eval, Context, OutputFormat};
-use error::CliError;
 use parser::{parse, Ast};
-
-mod error;
-mod package;
 
 #[derive(Parser, Debug)]
 #[command(author, version, about)]
@@ -76,8 +75,8 @@ fn print_tree(tree: &Ast) {
     println!("{}", tree.tree_string());
 }
 
-fn watch(args: &Args, target: &String) -> Result<(), CliError> {
-    fn watch_compile(
+async fn watch(args: &Args, target: &String) -> Result<(), CliError> {
+    async fn watch_compile(
         event: notify::Result<Event>,
         args: &Args,
         target: &String,
@@ -140,22 +139,23 @@ fn watch(args: &Args, target: &String) -> Result<(), CliError> {
 
     watcher.watch(Path::new(&args.input), RecursiveMode::Recursive)?;
 
-    watch_compile(Ok(Event::default()), args, target)?;
+    watch_compile(Ok(Event::default()), args, target).await?;
 
     for event in rx {
-        watch_compile(event, args, target)?;
+        watch_compile(event, args, target).await?;
     }
 
     Ok(())
 }
 
-fn main() -> Result<(), CliError> {
+#[tokio::main]
+async fn main() -> Result<(), CliError> {
     let args = Args::parse();
     let current_path = env::current_dir()?;
     let target = current_path.into_os_string().into_string().unwrap();
 
     if args.watch {
-        watch(&args, &target)?;
+        watch(&args, &target).await?;
     } else {
         match compile_file(&args) {
             Ok((tree, state)) => {
@@ -195,5 +195,9 @@ fn main() -> Result<(), CliError> {
         }
     }
 
+    /*
+        let pm = package::PackageManager;
+        let wasm_module = pm.resolve("pkgs:math").await?;
+    */
     Ok(())
 }
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -75,8 +75,8 @@ fn print_tree(tree: &Ast) {
     println!("{}", tree.tree_string());
 }
 
-async fn watch(args: &Args, target: &String) -> Result<(), CliError> {
-    async fn watch_compile(
+fn watch(args: &Args, target: &String) -> Result<(), CliError> {
+    fn watch_compile(
         event: notify::Result<Event>,
         args: &Args,
         target: &String,
@@ -139,23 +139,22 @@ async fn watch(args: &Args, target: &String) -> Result<(), CliError> {
 
     watcher.watch(Path::new(&args.input), RecursiveMode::Recursive)?;
 
-    watch_compile(Ok(Event::default()), args, target).await?;
+    watch_compile(Ok(Event::default()), args, target)?;
 
     for event in rx {
-        watch_compile(event, args, target).await?;
+        watch_compile(event, args, target)?;
     }
 
     Ok(())
 }
 
-#[tokio::main]
-async fn main() -> Result<(), CliError> {
+fn main() -> Result<(), CliError> {
     let args = Args::parse();
     let current_path = env::current_dir()?;
     let target = current_path.into_os_string().into_string().unwrap();
 
     if args.watch {
-        watch(&args, &target).await?;
+        watch(&args, &target)?;
     } else {
         match compile_file(&args) {
             Ok((tree, state)) => {
@@ -195,9 +194,5 @@ async fn main() -> Result<(), CliError> {
         }
     }
 
-    /*
-        let pm = package::PackageManager;
-        let wasm_module = pm.resolve("pkgs:math").await?;
-    */
     Ok(())
 }
--- a/cli/src/package.rs
+++ b/cli/src/package.rs
@@ -1,36 +1,115 @@
-use std::{env::current_dir, fs, path::PathBuf};
-
+use crate::error::CliError;
+use async_trait::async_trait;
 use directories::ProjectDirs;
+use serde_json;
+use std::{
+    env::current_dir,
+    fs::{self, create_dir_all, File},
+    io::copy,
+    path::PathBuf,
+};
 
+#[async_trait]
 pub trait Resolve {
-    fn resolve(&self, path: &str) -> Option<Vec<u8>>;
+    async fn resolve(&self, path: &str) -> Result<Vec<u8>, CliError>;
 }
 
 pub struct PackageManager;
+
+#[async_trait]
 impl Resolve for PackageManager {
-    fn resolve(&self, path: &str) -> Option<Vec<u8>> {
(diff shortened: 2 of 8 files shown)


cli/src/package.rs
```diff
@@ -1,36 +1,115 @@
...
+use serde_json;
+use std::{
+    env::current_dir,
+    fs::{self, create_dir_all, File},
+    io::copy,
+    path::PathBuf,
+};
 
+#[async_trait]
 pub trait Resolve {
```
adelhult, line 13: Move to Core and maybe remove `async`, you could just block before sending it to the resolve function. Also add a `resolve_all` function that resolves multiple paths in parallel

cli/src/main.rs
```diff
@@ -195,5 +195,9 @@ fn main() -> Result<(), CliError> {
         }
     }
 
+    /*
```
adelhult, line 198: Stray comment

adelhult (approved):
Nice work, the changes looks good!

CMDJojo (commented):
Looks good! Just saw a thing I thought I can comment on

cli/src/package.rs
```diff
@@ -1,36 +1,144 @@
...
+            .block_on(resolve_packages(paths))?;
+
+        return Ok(out);
+    }
+}
+
+async fn resolve_packages(paths: Vec<&str>) -> Result<Vec<Vec<u8>>, CliError> {
+    let futures = paths.iter().map(|path| resolve_package(path));
+
+    let results = join_all(futures).await;
```
CMDJojo, line 40: I think this could be `join_all(futures).await.collect::<Vec<Result<Vec<u8>, CliError>>().into()` and return immediately.

CMDJojo:
This also resolves GH-52 since it renames `core` to `modmark_core`, updated the description of this PR to reflect that

cli/src/package.rs
```diff
@@ -1,36 +1,136 @@
...
 
-        let mut cache_path = match ProjectDirs::from("org", "modmark", "packages") {
-            Some(path) => path.cache_dir().to_path_buf(),
-            None => return None,
+    let Some((specifier, package_path)) = splitter else {
+           return fetch_local(path);
         };
 
-        cache_path.push(PathBuf::from(path));
+    match specifier {
```
CMDJojo, line 50: Since this is the last statement of the body and `match`es in rust are actually  expressions, we can remove `return` from each branch of the match and just let Rust implicitly return the value

cli/src/package.rs
```diff
@@ -1,36 +1,136 @@
...
+        Some(path) => path.cache_dir().to_path_buf(),
+        None => return Err(CliError::Cache),
+    };
+
+    cache_path.push(PathBuf::from(&package_path));
+
+    let mut path = cache_path.clone();
+    path.pop();
+
+    create_dir_all(PathBuf::from(&path))?;
```
CMDJojo, line 68: This `let mut path ... path.pop ... pathbuf::from ...` thing to get the parent can be replaced with just `path.parent()` when using it, and here is the only spot you are using it,

cli/src/package.rs
```diff
@@ -1,36 +1,136 @@
...
-pub struct PackageManager;
 impl Resolve for PackageManager {
-    fn resolve(&self, path: &str) -> Option<Vec<u8>> {
-        let mut local_path = match current_dir() {
-            Ok(p) => p,
-            Err(_) => return None,
-        };
+    type Error = CliError;
+    fn resolve(&self, path: &str) -> Result<Vec<u8>, Self::Error> {
+        let out = tokio::runtime::Builder::new_multi_thread()
```
CMDJojo, line 18: We might want to make the whole package manager into a struct which we use as a singleton, which starts a tokio runtime the first time we actually run it, so we don't make new runtimes all the time (which actually creates new OS threads, one per CPU core etc). Also, we might not need to spin up a mutli-thread runtime if we only want to resolve one package, since yeah...it is only one job. Maybe add `.worker_threads(1)`? Or, if we have a singleton which owns its runtime, we may just schedule it there..

cli/src/package.rs
```diff
@@ -1,36 +1,136 @@
...
+            .enable_all()
+            .build()
+            .unwrap()
+            .block_on(resolve_package(path))?;
+
+        return Ok(out);
+    }
+    fn resolve_all(&self, paths: Vec<&str>) -> Vec<Result<Vec<u8>, Self::Error>> {
+        let out = tokio::runtime::Builder::new_multi_thread()
```
CMDJojo, line 27: This spins up multiple os threads (one per CPU core) each time we call any resolve_all. We might want to create the runtime only once.

CMDJojo (changes requested):
Good job with the package manager! I think it is a good start, and even though we might want to do some changes later on, I think those will be more apparent when we develop the language further and work on actual import statements. However, I think there are some things we would want to fix before merging. See comments.

cli/src/package.rs
```diff
@@ -1,36 +1,135 @@
...
-pub struct PackageManager;
 impl Resolve for PackageManager {
-    fn resolve(&self, path: &str) -> Option<Vec<u8>> {
-        let mut local_path = match current_dir() {
-            Ok(p) => p,
-            Err(_) => return None,
-        };
+    type Error = CliError;
+    fn resolve(&self, path: &str) -> Result<Vec<u8>, Self::Error> {
+        let out = RUNTIME.block_on(resolve_package(path))?;
```
CMDJojo, line 25: We don't need to bind this to a variable, we can return it directly
```suggestion
        RUNTIME.block_on(resolve_package(path))
```

cli/src/package.rs
```diff
@@ -1,36 +1,135 @@
...
+
+        return Ok(out);
+    }
+    fn resolve_all(&self, paths: Vec<&str>) -> Vec<Result<Vec<u8>, Self::Error>> {
+        let out = RUNTIME.block_on(resolve_packages(paths));
```
CMDJojo, line 30: Also here, we can return it directly
```suggestion
        RUNTIME.block_on(resolve_packages(paths))
```

cli/src/package.rs
```diff
@@ -1,36 +1,135 @@
...
-        cache_path.push(PathBuf::from(path));
+    match specifier {
+        "http" => fetch_url("", path).await,
+        "https" => fetch_url("", path).await,
+        "pkgs" => fetch_registry(package_path).await,
+        other => Err(CliError::Specifier(other.to_string())),
+    }
+}
+
+async fn fetch_url(url_root: &str, package_path: &str) -> Result<Vec<u8>, CliError> {
```
CMDJojo, line 57: (Maybe there is a good reason for this but) what is the reason for having two arguments "url_root" and "package_path"? What do they mean? Why is "url_root" always empty? The function is only called in two places. Maybe there is a good reason, plans for future work, please tell :D

cli/src/package.rs
```diff
@@ -1,36 +1,135 @@
...
+    let mut cache_path = match ProjectDirs::from("org", "modmark", "packages") {
+        Some(path) => path.cache_dir().to_path_buf(),
+        None => return Err(CliError::Cache),
+    };
+
+    cache_path.push(PathBuf::from(&package_path));
```
CMDJojo, line 63: Hmm, this means that (I have tested this) loading the url `https://github.com/modmark-org/packages/raw/main/bold.wasm` will make a cache file in `/Users/cmdjojo/Library/Caches/org.modmark.packages/https://github.com/modmark-org/packages/raw/main/bold.wasm` which I don't think is intentional; I don't think `https://` is a valid dir name and it causes confusion for MacOS. I don't think the `https://` is needed, maybe just the domain and path?

cli/src/package.rs
```diff
@@ -1,36 +1,135 @@
...
-            Some(path) => path.cache_dir().to_path_buf(),
-            None => return None,
+async fn resolve_packages(paths: Vec<&str>) -> Vec<Result<Vec<u8>, CliError>> {
+    let futures = paths.iter().map(|path| resolve_package(path));
+
+    return join_all(futures).await;
+}
+
+async fn resolve_package(path: &str) -> Result<Vec<u8>, CliError> {
+    let splitter = path.split_once(":");
```
CMDJojo, line 43: You can change this to split on a char instead, for better performance.
```suggestion
    let splitter = path.split_once(':');
```

CMDJojo (changes requested):
Good job! The package manager looks even better now! However, there is one fatal problem; loading "https://github.com/modmark-org/packages/raw/main/bold.wasm" will create some cache under "/Users/cmdjojo/Library/Caches/org.modmark.packages/https://github.com/modmark-org/packages/raw/main/bold.wasm" which I don't think is intentional. But it seems to be very close to being done now

cli/src/package.rs
hugomardbrink: Plan for when we add future support for different suffixes, GH:package_name.

CMDJojo: That sounds great!

cli/src/package.rs
hugomardbrink: Yeah I agree

cli/src/package.rs
```diff
@@ -1,36 +1,138 @@
...
-pub struct PackageManager;
 impl Resolve for PackageManager {
-    fn resolve(&self, path: &str) -> Option<Vec<u8>> {
-        let mut local_path = match current_dir() {
-            Ok(p) => p,
-            Err(_) => return None,
-        };
+    type Error = CliError;
+    fn resolve(&self, path: &str) -> Result<Vec<u8>, Self::Error> {
+        Ok(RUNTIME.block_on(resolve_package(path))?)
```
CMDJojo, line 25: You won't need the `Ok(...?)` wrapper here

cli/src/package.rs
```diff
@@ -1,36 +1,138 @@
...
+        Some(path) => path.cache_dir().to_path_buf(),
+        None => return Err(CliError::Cache),
+    };
+
+    let splitter = package_path.split_once(':');
+    let Some((_, mut package_path)) = splitter else {
+        return Err(CliError::Cache)
+    };
+
+    package_path = &package_path[2..];
```
CMDJojo, line 64: This crashes on giving the url `http:`

cli/src/package.rs
```diff
@@ -1,36 +1,138 @@
...
 
-        if local_path.exists() {
-            fs::read(local_path).ok();
-        }
+    if cache_path.exists() {
+        Ok(fs::read(cache_path)?)
+    } else {
+        println!("{}", package_path);
+        let response = reqwest::get(package_path).await?;
+        let content = response.text().await?;
```
CMDJojo, line 79: This should really be `response.data()` since we are capturing data, not text, and we want it to not try to pick up text encoding etc. Also, we need to check for success status code, since if you get a 404 page, you may get some text back even though we don't want to use the result (since it is failed)

cli/src/package.rs
```diff
@@ -1,36 +1,138 @@
...
+    create_dir_all(path)?;
 
-        if local_path.exists() {
-            fs::read(local_path).ok();
-        }
+    if cache_path.exists() {
+        Ok(fs::read(cache_path)?)
+    } else {
+        println!("{}", package_path);
+        let response = reqwest::get(package_path).await?;
```
CMDJojo, line 78: This has the wrong URL, you rewrite the package path when calculating the cache dir and because of that, we don't get `https://github.com/...`, rather `github.com/...` and this fails

CMDJojo (changes requested):
Good job! We are almost done. Just some small bugs I found that might have gotten introduced when you changed my previous comments haha

CMDJojo (approved):
Nice work with the PM! It will be nice to get this merged to continue working on #169 to actually make stuff importable. Ready to merge!

PR #129, opened 2023-03-02
GH-62: Add layout package

Forsinge:
Resolves: #62 

This PR adds the layout package, which supports both row modules and center modules.


packages/layout/src/main.rs
```diff
@@ -0,0 +1,180 @@
...
+        "name": "layout",
+        "version": "0.1",
+        "description": "This package provides primitive layout modules.",
+        "transforms": [
+            {
+                "from": "row",
+                "to": ["html"],
+                "arguments": [
+                    {"name": "separator", "default": ",", "description": "The pattern used to separate items in the input content." },
+                    {"name": "gap", "default": "10", "description": "The gap between items given in pixels." },
```
CMDJojo, line 31: Hmm, CSS conventions are moving from absolute units as pixels to rem. We might want to be able to allow for using `rem` instead of `px`, or allow them to add an extension, like `10px` but use `rem` by default.

packages/layout/src/main.rs
```diff
@@ -0,0 +1,180 @@
...
+                    {"name": "max_width", "default": "none", "description":
```
CMDJojo, line 32: Same here, maybe change to rem

packages/layout/src/main.rs
```diff
@@ -0,0 +1,180 @@
...
+            eprintln!("Unexpected value for argument: max_width")
+        }
+    }
+
+    if wrap == "true" {
+        write!(style, "flex-wrap: wrap; ").unwrap();
+    } else {
+        write!(style, "overflow: hidden; ").unwrap();
+        if wrap != "false" {
+            eprintln!("Unexpected value for argument: wrap")
```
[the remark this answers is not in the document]
CMDJojo: (But maybe add the text that it expected true/false, here and at the other places you use it)

CMDJojo (changes requested):
Nice, really good work! It will be nice to get some layout things in HTML (and we might want to add even more later on). I have some small things to comment on, but no biggies, they will be quick fixes. It would also be nice with some simple tests!

Forsinge:
The most recent commit should address all your comments, with the addition of your idea to allow the user to specify any css unit. I ended up having to change quite a bit even for such small changes, so I'd appreciate if you could take another look!

packages/layout/src/main.rs
```diff
@@ -0,0 +1,200 @@
...
+            "transforms": [
+                {
+                    "from": "row",
+                    "to": ["html"],
+                    "arguments": [
+                        {"name": "separator", "default": ",", "description": "The pattern used to separate items in the input content." },
+                        {"name": "gap", "default": "10", "description":
+                            "The gap between items. You can \
+                            optionally add a css unit, otherwise \
+                            rem will be used.."
```
CMDJojo, line 36: Two dots, typo

CMDJojo (approved):
Really nice work, good job! Looks fantastic now, the units was a nice addition. Ready to merge (if you don't want to fix the typo I just mentioned)

packages/layout/src/main.rs
Forsinge: Fixed!

Issue #130, opened 2023-03-02
Add a open in overleaf button to the playground


bruhng:
**Is your feature request related to a problem? Please describe.**
It is annoying to copy paste the latex code to see if it compiles, this would make that a lot easier

**Describe the solution you'd like**
Add a button that calls the overleaf api that opens the compiled code in overleaf

**Additional context**
How to implement:
https://www.overleaf.com/devs

PR #138, opened 2023-03-03
GH-130: Add a open in overleaf button in the playground

bruhng:
- Resolves GH-130:
Adds a button in the playground that opens the compiled code in overleaf


CMDJojo (changes requested):
It is nice to have an "Open in Overleaf" button, nice idea of you! Also, nice clever solution to send post data using a hidden form. However, I think there are some problems; first of all, the button is visible all the time, even when it doesn't make sense. We might want to add a custom "Output in LaTeX" in the drop-down menu, or just make sure that it doesn't show if we haven't chosen "latex" as the output language. Additionally, it has a kind-of awkward placement in relation to the other buttons. Lastly, it still open Overleaf but with some invalid data which gives a big red error, if the LaTeX output isn't chosen. I think this can be improved greatly.

playground/static/main.js
```diff
@@ -238,6 +238,8 @@ async function updateOutput(input) {
                 renderIframe.style.display = "none";
                 render.style.display = "none";
```
CMDJojo, line 240: Maybe remove this newline to improve style, so that all "set style"-statements stick together

playground/static/main.js
```diff
@@ -247,6 +249,8 @@ async function updateOutput(input) {
...
```
CMDJojo, line 251: Maybe remove this newline to improve style, so that all "set style"-statements stick together

playground/static/main.js
```diff
@@ -256,6 +260,8 @@ async function updateOutput(input) {
...
```
CMDJojo, line 262: Maybe remove this newline to improve style, so that all "set style"-statements stick together

playground/static/main.js
```diff
@@ -265,22 +271,43 @@ async function updateOutput(input) {
                 errors.forEach(addError);
                 warnings.forEach(addWarning);
 
+                overleafButton.style.display = "none";
+
                 debugEditor.container.style.display = "block";
                 renderIframe.style.display = "none";
                 render.style.display = "none";
```
CMDJojo, line 279: Maybe remove this newline to improve style, so that all "set style"-statements stick together

playground/static/main.js
```diff
@@ -265,22 +271,43 @@ async function updateOutput(input) {
                 errors.forEach(addError);
                 warnings.forEach(addWarning);
 
+                overleafButton.style.display = "none";
```
CMDJojo, line 274: Maybe remove this line altogether, since we probably don't need two duplicate statements in the same block

playground/static/main.js
```diff
@@ -265,22 +271,43 @@ async function updateOutput(input) {
...
                 debugEditor.getSession().selection.clearSelection();
             }
                 break;
+            case "latex":
+                let { content, warnings, errors } = JSON.parse(await compilerAction({ type: "transpile", source: input, format: "latex" }));
+                errors.forEach(addError);
+                warnings.forEach(addWarning);
+
+                overleafButton.style.display = "inline-block";
+
```
CMDJojo, line 293: Maybe remove this newline to improve style, so that all "set style"-statements stick together

playground/static/main.js
```diff
@@ -265,22 +271,43 @@ async function updateOutput(input) {
...
+                debugEditor.container.style.display = "block";
+                renderIframe.style.display = "none";
+                render.style.display = "none";
+                
+                debugEditor.session.setMode("");
```
CMDJojo, line 298: Maybe import LaTeX style?

CMDJojo (changes requested):
It looks much better now! Also nice to have a specific LaTeX output tab, especially when we are targeting LaTeX with all packages we are making ourselves. There are some style changes I would like to see fixed before merging, but it is looking much better now!

Issue #135, opened 2023-03-03
Cannot use " when outputting to LaTeX


CMDJojo:
**Describe the bug**
You can't use a quotation mark `"` when outputting to LaTeX. If writing a quotation mark which isn't replaced by a smart punctuation character, and the LaTeX thus gets the character `"` as input, it fails with the error `__text -> latex: Error deserializing result from module`.

**To Reproduce**
1. Go to the playground and set the output language to LaTeX
2. Enter any text that supplies the `"` character to the LaTeX package. The minimum example is just the character `"`

**Expected behaviour**
That the module shouldn't fail, and that we should get an output

**Screenshots**
<img width="1086" alt="Screenshot 2023-03-03 at 10 37 04" src="https://user-images.githubusercontent.com/25929684/222685660-a3a7b082-4b24-477b-a174-e0c4a1c19cae.png">

PR #139, opened 2023-03-03
GH-135: Fix bugg where you can not use a single quote

bruhng:
- Resolves GH-135:
Fixes the bug where you can not use a single "

### DIFF
--- a/packages/latex/src/main.rs
+++ b/packages/latex/src/main.rs
@@ -156,6 +156,7 @@ fn escape_text(module: Value) -> String {
         let s = s.split('\\').map(|t| t.replace('{', r"\\{").replace('}', r"\\}")).collect::<Vec<String>>().join(r"\\textbacklash{}")
             .replace('#', r"\\#")
             .replace('$', r"\\$")
+            .replace("\"","\\\"")
             .replace('%', r"\\%")
             .replace('&', r"\\&")
             .replace('_', r"\\_")


packages/latex/src/main.rs
```diff
@@ -156,6 +156,7 @@ fn escape_text(module: Value) -> String {
         let s = s.split('\\').map(|t| t.replace('{', r"\\{").replace('}', r"\\}")).collect::<Vec<String>>().join(r"\\textbacklash{}")
             .replace('#', r"\\#")
             .replace('$', r"\\$")
+            .replace("\"","\\\"")
```
CMDJojo, line 159: This makes it seem like you are escaping the LaTeX character, which isn't what you actually are doing. You are adding a backslash behind the quote to escape the JSON formatting, which doesn't have anything to do with escaping " in LaTeX. See line 168, that's where the bug originates from. You are just inserting the string `s` into the JSON, which makes you need to escape all backslashes in this and all other replacements. I think this is bad and that it would be much better to use serde_json to serialize the JSON, so it is clear what you are doing so others understand easier and don't make the same mistake. You can, for example, use the json! macro to generate the json for you. Also, you having to replace each newline with a space is also due to the json not being serialized properly, and fixing that would make the output more alike the input, and look better.

CMDJojo (changes requested):
I think you should look at line 168 and fix the source of the error by properly serializing the json data. This means that you don't need to replace newlines and such.

CMDJojo (changes requested):
The changes look much better now, now it is much more clear what you are escaping and when, and why etc. However, you have some things to fix before merging:
- ***Add tests***, there are currently **no tests at all** for the `Latex` package, and it would be nice to have some to verify that bugs like this one isn't left in the source code, and to make sure regression doesn't happen.
- Rebase with `main`; this package can't be merged into main since it has conflicts, and furthermore violates the package conventions introduced in #143 and enforced in #144. You will need to rebase with main and make sure that the tests pass.

Issue #140, opened 2023-03-06
Unintended output from inline code modules


Forsinge:
**Describe the bug**
The code package uses `<pre>` elements to display code snippets. These are block-level elements which will force an open `<p>` to close. Therefore, when the code module is used inline surrounding paragraph tags are closed separately, resulting in `<p></p><pre>...</pre><p></p>`.

**To Reproduce**
Type `[code]abc` in the playground and select html output.

PR #141, opened 2023-03-06
GH-140: Fix HTML output from inline code modules

Forsinge:
Resolves: #140 

Note that I chose to omit giving warnings for inline usage. The changes of interest should be line 88 to 95, the rest was done by ``cargo fmt``.

### DIFF
--- a/packages/code/src/example.json
+++ b/packages/code/src/example.json
@@ -1,6 +1,0 @@
-{
-  "name": "code",
-  "arguments": {},
-  "data": "todo",
-  "inline": false
-}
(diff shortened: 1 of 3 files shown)


CMDJojo (approved):
Looks good to me! Nice to get it working correctly inline. Also, nice catch of you that it adds lots of `<p>` tags. I am not good at HTML enough to figure that out.

CMDJojo:
Might be worth adding a test using it inline, to make sure that it works as expected, but that's no requirement

CMDJojo:
Screenshot 2023-03-08 at 16 09 35
Also, just found this out, the boxes are quite tall and collides with other text

Forsinge:
Good catch. I added a test and reduced the padding to 0.1 rem, seemed fine to me.

PR #144, opened 2023-03-06
GH-143: Clarify and enforce package conventions

### LINKED ISSUE #143: Clarify and enforce package conventions
Currently, we are developing packages but we have little to none written package conventions. This looks a little bad, especially when inspecting the packages in the playground, when there is one package called "Standard table package" with spaces, uppercase letters, one called "Latex" which has uppercase letters but no spaces and one called "core" with only lowercase letters.

I would like to streamline the packages by using these conventions:
* For our packages in cargo crates under `./packages`, they must be in a folder named the same thing as the crate (currently, all are, and imports wouldn't work especially good without that)
    * In addition, the crate must have exactly (1) binary target, named the same thing as the crate and folder
* Package names must start with a lowercase letter and may contain lowercase letters, underscores and dashes.
    * This will allow for case-insensitive searching for packages (we can just make the name lowercase before a lookup), and will probably lead to developers naming their packages the same as the main module contained within, making it easier to see the connection between modules and packages. It will be error-prone if `[import] table` and `[import] Table` meant different things, especially since the users may not have good coding knowledge, and `[import] "Standard table module"` looks bad.
    * Underscores and dashes allows for `table_supercharged`, and numbers allows for `tablev3`. Maybe suggest only using underscores but allow dashes?
* Each package must contain at least one transform
    * (A package isn't very useful without one)
* Each transform must satisfy the following:
    * Its `from` field (the parent or module name) must only contain lowercase letters, digits, dashes, underscores and dots
        * Dots is a futureproofing if we want to namespace. We might want to allow uppercase to do `Tag.Bold`, or just go for `tag.bold`
    * Its `to` array must have at least one entry and they all must be in all lowercase letters
        * Since this is meant to match a format (or an extension), this seems reasonable
    * Its arguments should have names starting with a lowercase letter and may only contain lowercase letters, numbers and underscores
        * The parser only allows alphanumerics and underscores, so arguments not following this won't be able to be invoked by name
        * Maybe we want to drop the limitation that it has to start with a letter, but I think it makes sense to have that restriction

We could not have any conventions, but I think the end user experience won't be very streamlined. We could change some restrictions, like making arguments case-sensitive, but I think that would be confusing and I don't see any benefits with that. Currently, the parser allows multi-cased module names and argument names, so I think they could be converted to lower-cased after parsing.

CMDJojo:
See issue #143. This PR resolves that issue by adding the conventions discussed there into `packages/README.md` and adds a test in `package-tests` which checks that these conventions are being met. Additionally, the convention test gives a warning if a package doesn't have any tests in it (but it doesn't fail the test). This PR also changes the current packages to follow the convention.

Resolves #143

### DIFF
--- a/package-tests/tests/package_tests.rs
+++ b/package-tests/tests/package_tests.rs
@@ -7,6 +7,7 @@ use std::sync::{Arc, Mutex};
 
 use diffy::create_patch;
 use once_cell::sync::Lazy;
+use serde::{Deserialize, Serialize};
 use serde_json::Value;
 
 // Unfortunately, if we run multiple tests at the very same time, some of them will sometimes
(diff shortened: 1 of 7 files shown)


package-tests/tests/package_tests.rs
```diff
@@ -105,8 +106,205 @@ fn test_package_input(file: &Path) -> datatest_stable::Result<()> {
...
+
+    // Now, let's check the package manifest
+    // Check that the name is lowercase/digit/_/- and starts lowercase
+    let name = manifest.name;
+    let ok_name = name.starts_with(|c: char| c.is_ascii_lowercase())
+        && name
+            .chars()
+            .all(|c| c.is_ascii_lowercase() || c.is_ascii_digit() || c == '_' || c == '-');
+
+    assert!(ok_name, "Package name must start with lowercase letter, and must only contain lowercase letters, digits, underscores and dashes");
```
Forsinge, line 211: Maybe it could also check that the name provided by the package manifest matches the others.

Forsinge (changes requested):
Looks good! Good comments in the code as well, makes it easier to follow.

I found one thing that could be looked into, should be in a comment below. I also noticed the "Blocking ..." messages clutter the output a lot during tests, but I take it that's not a direct result of your code.

Given that the packages follow a naming convention now, maybe we can also sort the packages alphabetically in the playground?

Forsinge (approved):
LGTM!

PR #148, opened 2023-03-09
GH-145: Fix inconsistent error on missing transform

### LINKED ISSUE #125: CoreError when previewing JSON if an argument is supplied to a non-existing module
**Describe the bug**
There is a "JSON tree" output option for the playground, which is useful to see the JSON representation for the document. This may be used to get a valid input to a package, and is useful for package development. 

If one writes a module invocation like `[module] body`, we can see the JSON representation for this call (even though the module doesn't exist, since the document isn't rendered, just parsed and converted to JSON), but it doesn't work if we supply arguments to that module.

**To Reproduce**
Steps to reproduce the behaviour:
1. Go to the playground and change the output to JSON tree
2. Enter `[module] body` and notice how you get a JSON tree (see first screenshot)
3. Enter `[module arg1=foo] body` and notice how the whole thing errors out and we don't get a JSON tree (see second screenshot)

**Expected behaviour**
I would expect that we, in both cases, would get a JSON tree. The expected output is:
```json
{
  "name": "__document",
  "arguments": {},
  "children": [
    {
      "name": "__paragraph",
      "arguments": {},
      "children": [
        {
          "name": "module",
          "data": "body",
          "arguments": {"arg1": "foo"},
          "inline": true
        }
      ]
    }
  ]
}
```

**Screenshots**
Screenshot 2023-03-01 at 17 04 58
Screenshot 2023-03-01 at 17 13 24

### LINKED ISSUE #145: Wrong type of error when invoking a non-existent module with arguments
**Describe the bug**
When trying to invoke a non-existent module or a module which doesn't support a transform to the target output language, and you supply no arguments to it, it says "No package for transforming node ... to ...", which is reasonable. This clearly tells the user what is wrong and why the document thus failed to compile. However, this does not work when supplying arguments to it, it says "Argument ... is not supported by the element ..." which isn't wrong per se, but isn't expected either.

**To Reproduce**
1. Write `[module] body` (assuming no module named `module` exists) and notice the error you get
2. Change this to `[module args] body` and notice that you get a different error, which isn't the same as before

**Expected behavior**
The expected behaviour is to get the same (or a similar error) "No package for transforming node ... to ..." since that is more appropriate than  "Argument ... is not supported by the element ...".

**Screenshots**
image
image

**Additional context**
Related to #125

CMDJojo:
This PR fixes the inconsistent error type when you have a missing transform. #145 outlines how an unknown module expression without arguments would give an "No package for transforming..." error while making that same expression with arguments would give another error. This PR fixes it so that it always yields "No package for transforming..."-error regardless of if arguments are supplied or not.

Moreover, this touches on #125 since the error originated from that it was possible to serialise JSON for non-existing modules, assuming they had no arguments. The example in #125 will no longer produce valid JSON, since Core would impossibly know what to name unnamed arguments when constructing the map. #125 will thus have an outcome other than the one suggested, but the inconsistencies will be gone. One alternative, to get an JSON representation for modules that do not exist, would be to make up names for their arguments but I find that deeply unsatisfactory since it would require workarounds just to make that output option work for invalid input.

- Resolves #145 
- Resolves #125

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -69,7 +69,7 @@ impl CompilationState {
 }
 
 impl Debug for Context {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
         f.debug_struct("Context")
             .field("packages", &self.packages)
             .field("transforms", &self.transforms)
@@ -218,6 +218,7 @@ impl Context {
         Ok(())
     }
 
+    //noinspection RsLiveness
     /// This is a helper function to load a package directly from its wasm source. It will be
     /// compiled using `Package::new` to become a `Package` and then loaded using `load_package`
     pub fn load_package_from_wasm(&mut self, wasm_source: &[u8]) -> Result<(), CoreError> {
@@ -230,6 +231,18 @@ impl Context {
         self.load_package(pkg)
     }
 
+    /// Gets the transform and package the transform is in, for a transform from a specific element
+    /// to a specific output format. Returns None if no such transform exists
+    pub fn get_transform_to(
+        &self,
+        element_name: &str,
+        output_format: &OutputFormat,
+    ) -> Option<&(Transform, Package)> {
+        self.transforms
+            .get(element_name)
+            .and_then(|t| t.find_transform_to(output_format))
+    }
+
     /// This gets the transform info for a given element and output format. If a native package
     /// supplies a transform for that element, that will be returned and the output format returned
     pub fn get_transform_info(
@@ -237,9 +250,7 @@ impl Context {
         element_name: &str,
         output_format: &OutputFormat,
     ) -> Option<&Transform> {
-        self.transforms
-            .get(element_name)
-            .and_then(|t| t.find_transform_to(output_format))
+        self.get_transform_to(element_name, output_format)
             .map(|(transform, _)| transform)
     }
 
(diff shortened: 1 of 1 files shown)


core/src/context.rs
```diff
@@ -333,8 +344,10 @@ impl Context {
         let mut err_out = Pipe::new();
 
         // Generate the input data (by serializing elements)
+        println!("Before ser");
```
adelhult, line 347: debug printing

core/src/context.rs
```diff
@@ -333,8 +344,10 @@ impl Context {
...
         let input_data = self.serialize_element(from, output_format)?;
         write!(&mut input, "{}", input_data)?;
+        println!("After ser");
```
adelhult, line 350: debug printing

core/src/context.rs
```diff
@@ -17,6 +17,7 @@ use wasmer_wasi::{Pipe, WasiState};
 use parser::ModuleArguments;
 
 use crate::package::PackageImplementation;
+use crate::CoreError::MissingTransform;
```
adelhult, line 20: I think that we don't need to bring the enum variant into scope and instead use it as `CoreError::MissingTransform` like we do with all other errors.

adelhult (approved):
Nice of you to fix a few rough edges! Left a few comments.

core/src/context.rs
CMDJojo: Good catch! Fixed!

core/src/context.rs
CMDJojo: Fixed!

core/src/context.rs
CMDJojo: In my perfect world, all enums would be in scope all the time :( but okay, I'll remove it

CMDJojo:
Force-pushed with fixes to the comments

PR #150, opened 2023-03-10
GH-94: Improve table package

CMDJojo:
This PR aims to resolve GH-94 by adding wanted improvements to the Table package. Here are the points listed, with my comments:

- [x] Fix quote and backslash support
        - *Works as expected, one backslash escapes the delimiter, two doesn't, three does etc*
- [x] Arguments for alignment
        - *There is an `alignment` argument which you can set to left/center/right or some combination of l/c/r, one character for each column, like ccr which would center the two first columns and right-align the third.*
- [x] Arguments for borders
        - *Argument `borders` can be `all`/`horizontal`/`vertical`/`outer`/`none`. The options are self-explanatory. All but `none` implies `outer`*
- [x] Option for first row to be headers
        - *Argument `header` can be `none` which has no special treatment or `bold` which is `th` in html and `textbf` in latex*
- [x] Strip whitespace
        - *Argument `strip_whitespace` (true/false) controls stripping of whitespace*

~~This is a draft PR for now, and these are the things that needs to be fixed before promoting to a "real" PR:~~

- [x] Commenting code
- [x] Adding tests
- [x] Make sure it works for empty tables (for LaTeX, we insert `|l|` as the alignment for empty tables since it errors without an alignment. Also, we give an warning)
- [x] Possibly functional decomposition, the code doesn't look fantastic right now (I did that somewhat, think it got somewhat better)
- [x] General clean-up

Since the above-mentioned tasks are fixed, this PR is ready for review!

- Resolves GH-94
- Resolves GH-112

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -127,12 +127,30 @@ mod tests {
             description: "This package supports [table] modules".to_string(),
             transforms: vec![Transform {
                 from: "table".to_string(),
-                to: vec![OutputFormat::new("html")],
-                arguments: vec![ArgInfo {
-                    name: "col_delimiter".to_string(),
-                    default: Some("|".to_string()),
-                    description: "The string delimiter for columns".to_string(),
-                }],
+                to: vec![OutputFormat::new("html"), OutputFormat::new("latex")],
+                arguments: vec![
+                    ArgInfo {
+                        name: "header".to_string(),
+                        default: Some("none".to_string()),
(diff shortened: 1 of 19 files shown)


CMDJojo:
(The reason for the previous test failures weren't the table module itself, rather that Core runs a test loading it and reading the manifest, and since I didn't update that test with the new manifest, it failed)

CMDJojo:
I think this PR is ready for review now!

adelhult (approved):

PR #151, opened 2023-03-10
Playground: Add chromium support

adelhult:
Resolves GH-45. This PR adds support for the playground in Chromium browsers. This is done by instantiating the WASM module that contains the compiler inside of a web worker instead of the main thread.

### DIFF
--- a/playground/static/main.js
+++ b/playground/static/main.js
@@ -1,5 +1,32 @@
-// noinspection JSFileReferences
-import init, {ast, ast_debug, json_output, package_info, transpile} from "./pkg/web_bindings.js";
+let seq = 0;
+let compiler_callback;
+let compiler_failure;
+let compiler = new Worker("./compiler.js");
+
+compiler.onmessage = (event) => {
+    // Render the document once the wasm module containing the
+    // compiler has been instantiated.
+    if (event.data.type === "init") {
+        updateOutput(editor.getValue());
+        return;
+    }
+
+    if (event.data.seq != seq) return;
+    if (event.data.success) {
+        compiler_callback(event.data.result);
+    } else {
+        compiler_failure(event.data.error);
+    }
+}
+
+async function compilerAction(action) {
+    let promise = new Promise((res, rej) => {
+        compiler_callback = res;
+        compiler_failure = rej;
+    });
+    compiler.postMessage({ seq: ++seq, ...action });
+    return promise;
+}
 
 let view = "editor";
 const editorView = document.getElementById("editor-view");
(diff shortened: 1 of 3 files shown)


CMDJojo:
Nice! Good job on the fix! However, I think we should change to import the script using ECMAScript module syntax since it looks cleaner, doesn't pollute the scope with stuff we don't know about, and have cleaner syntax overall when calling the functions. Firefox will support it in [version 111](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687), which is released [March 14th](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/111). This won't require changes to `wasm-pack` and so on, and is generally the preferred way to do it!

CMDJojo:
Also, this does not work...I think you failed to commit the `--no-modules` hack to get around ES module syntax. Doesn't run on FF110.0.1 on Windows 21H2

adelhult:
Yeah, noticed that too, forced push a change a few seconds ago.

adelhult:
> Nice! Good job on the fix! However, I think we should change to import the script using ECMAScript module syntax since it looks cleaner, doesn't pollute the scope with stuff we don't know about, and have cleaner syntax overall when calling the functions. Firefox will support it in [version 111](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687), which is released [March 14th](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/111). This won't require changes to `wasm-pack` and so on, and is generally the preferred way to do it!

Opened #152!

CMDJojo:
image
I would love it if you made it render on load again as it did before, as it is a nice welcoming screen when the example document gets loaded

adelhult:
Hmm, I thought I fixed that again. Will take a look!

playground/static/main.js
```diff
@@ -223,7 +245,7 @@ function updateOutput(input) {
                 render.style.display = "none";
 
                 debugEditor.session.setMode("");
-                debugEditor.setValue(ast_debug(input));
+                debugEditor.setValue(await compilerAction({ type: "ast", source: input }));
```
CMDJojo, line 251: This is probably wrong, it seems strange to do the action type `ast` on both `ast-debug` and `ast` arms, probably a typo. See line 239.

CMDJojo (changes requested):

CMDJojo:
image
I still get the same problem sometimes, I think we still have a race condition between the document fetch and the parser/core loading, possibly... It would be nice to get this resolved before pushing something that works sometimes. (FF110.0.1 on Windows 21H2)

CMDJojo:
We might be able to solve this by fetching/displaying the welcome document strictly after we have loaded the core wasm, by sending an init event to the main (sync) JS. Or let the worker fetch the document itself

CMDJojo:
It seems like your last commit solved the problem. I think we are ready to merge it, but we should PR #152 asap to merge it on tuesday when the Firefox update goes stable and we can verify that ES modules work there as well! (Approved)

CMDJojo (approved):

Issue #152, opened 2023-03-10
Use ES module for the web worker in playground 


adelhult:
**Describe the bug**
To support all major browser the web worker used to run the compiler in the playground currently uses a bit of an ugly import system. This is because ES modules (i.e. js that uses the nice `import` syntax) is not supported inside of web workers in Firefox. It looks like this is about to change in the next upcoming version, so we should switch to the cleaner solution then.

CMDJojo:
Current state: See pr #154. Firefox 111 has support for module workers but it is controlled by a flag which is disabled by default. It would probably be too much of an inconvenience for users to have to enable a flag to use our playground on a major browser, so I think we should shelve this until it has support without using a custom flag.

PR #154, opened 2023-03-12
GH-152: Use ES-module import syntax

CMDJojo:
This PR changes the `compiler.js` to import the bindings using the more modern ES-module syntax. This is supported by all major browsers but Firefox, which will get it in version 111, released [March 14th](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/111). After we confirm that this works, we should thus merge this PR.

It also changes `main.js` to have the extension `mjs` for JavaScript module, needed for support in MacOS.

Resolves GH-152

### DIFF
--- a/playground/build-playground.sh
+++ b/playground/build-playground.sh
@@ -1,4 +1,4 @@
 rm -r build
 mkdir build
 cp -r static/* build
-wasm-pack build ./web_bindings --target no-modules --out-dir ../build/pkg
+wasm-pack build ./web_bindings --target web --out-dir ../build/pkg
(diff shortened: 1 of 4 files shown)


CMDJojo:
This **does** work on Firefox 111.0 (beta) **but** is locked behind the `dom.workers.modules.enabled` flag (disabled by default). So, this will require users to change default settings to get it to work. I think we should hold off on changing this for the time being, until it actually has support without changing flags. Closing this PR for now.

CMDJojo:
Confirmed to still be locked behind that flag on Firefox 111.0 Stable Release

Issue #156, opened 2023-03-14
Make CLI not crash on core errors


CMDJojo:
**Is your feature request related to a problem? Please describe.**
When compiling a ModMark document, there are different errors that can occur. While some errors may occur within a module, in which case the ModMark compiler still runs and doesn't return an error itself, but rather returns some errors alongside the compiled document, some errors are CoreErrors which are returned as an `Err` before the document is compiled, and is in some sense more "critical". Some of these errors can be related to VM problems, but some are simply parse errors that occur when the user has specified unnamed arguments after named arguments. Currently, when running the Cli in the "watch" mode, it exits at the first error. This isn't very useful.

**Describe the solution you'd like**
I want to, upon a CoreError, print that error to the console and still keep the Cli running. This will make it more convenient to write documents in our language; malformed arguments isn't a reason for the Cli to stop the VM and have the user start it up again. See screenshot for a possible implementation:

<img width="475" alt="image" src="https://user-images.githubusercontent.com/25929684/224869355-bd8c2085-9208-45b4-b810-f1af2c3920d9.png">

PR #157, opened 2023-03-14
GH-156: Print CoreError in the Cli when running in Watch mode

CMDJojo:
This PR makes the Cli write out any CoreError when running with the `--watch` flag rather than terminating the program. See screenshot:

image

- Resolves GH-156


Forsinge (approved):
LGTM!

Issue #63, opened 2023-02-16
Package filesystem access


axoxelol1:
**Is your feature request related to a problem? Please describe.**
Packages need to be able to read files. For example a citation library needs access to a bibtex file or equivalent. 

**Describe the solution you'd like**
Wasmer allows mounting of sandboxed filesystems that could be useful.

PR #158, opened 2023-03-14
Add file access for packages

Forsinge:
Resolves: #63 

This PR enables packages to access files.

Features in browser:
- Uploading local files
- Creating folders
- Renaming folders and files (hover over list entries to see buttons)
- Removing folders and files
- Downloading files
- Navigating between folders (double click on a folder name!)

Features in cli:
- Access to a directory stored on host system
- Prompts for users to allow or deny access to a file for packages
- `--assets` or `-a` argument to specify a different path to the external directory
- `--deny-read` and `--deny-write` arguments to manage file access
- `--no-prompts` argument to disable the user prompts

Additionally it includes a very basic version of a "files" package. This supports transforms from [textfile], [include] and [image] to html.

Todo:
- [x] Improve frontend + add some consistency to the styling
- [x] Add downloads
- [x] Make design more responsive -- *result of this was not ideal, but should be sufficient for phone*
- [x] Handle and propagate errors
- [x] Handle duplicate names in filesystem
- [x] Reorganize and refactor (probably shouldn't be creating complicated html in web_bindings!)
- [x] Clean up imports and exports
- [x] Look into user prompts for file access
- [x] Add cli arguments for: filesystem permissions, specifying assets path, disabling user prompts
- [x] Add write and create permissions for preopens
- [x] Remove forgotten left-over code
- [x] General clean up -- *mostly done I think*


Forsinge:
I think this is ready for review now!

The files package is currently still included. I suppose there are two options: either we remove it from this pull request and start anew, or we include it and open a new issue to improve it.

core/src/native_fs.rs
```diff
@@ -0,0 +1,151 @@
...
+        self.inner.remove_file(path)
+    }
+
+    fn new_open_options(&self) -> OpenOptions {
+        OpenOptions::new(Box::new(self.file_opener.clone()))
+    }
+}
+
+#[derive(Clone)]
+pub struct NativeFileOpener {
```
adelhult, line 73: I thought that I could leave some input here regarding what we talked about earlier!

To decouple the file system from the system that handles file permissions and prompts the user for access we could create a new trait, maybe named `AccessPolicy` or something like that. This would work very much like the `Resolve` trait that is used by the package manager and be generic over the Context struct: `Context<SomeResolveImpl, SomeAccessPolicyImpl>`

Perhaps the trait could look something like this, although you definitely have more expertise in this area : )
```rs
trait AccessPolicy {
   allowed_to_read(path: &Path /*maybe include which package too*/) -> bool;
   allowed_to_write(path: &Path  /*maybe include which package too*/) -> bool;
}
```

Then we can write multiple implementations for this trait (one for the cli that uses the command-line flags and prompts the user if need be, very much like the code you have already written, and maybe one for the web that allows all operations)

adelhult (commented):
I think that everything looks really good! Left a comment about how we might go about generalizing the file access and move some more of the logic outside of the `core` crate

Forsinge:
The most recent commit does not include implementation of `Default` for `Context` (something we discussed briefly today) because `new()` returns a `Result`. We could unwrap to make it work, but this doesn't seem very appropriate. If you have other ideas let me know, otherwise I say we leave it unimplemented.

cli/src/file_access.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        print!("Unexpected input. Please enter (y/n): ");
+        stdout().flush().expect("Could not flush output");
+    }
+}
+
+impl CliAccessManager {
+    pub(crate) fn new(args: &Args) -> Self {
+        Self {
+            root: args.assets.clone().unwrap_or(String::from("assets")),
```
adelhult, line 82: If no `--assets` flag was specified we assume that the user wants to open a "assets" directory. I think that we instead should skip the preopen step and not open a directory at all. This way, we would be more secure by default and with the current system it is impossible to compile documents with exposing a directory:

If we compile without having a assets sub directory we get this error:
image

and we add  `--deny-read --deny-create --deny-write` all of these files even if we have a valid directory we get the following error:
 
image

adelhult (commented):
Great work :partying_face:

I only have one more suggestion before we are ready to rebase!

cli/src/file_access.rs
Forsinge: Fixed, I believe! I was unable to find a neat way to introduce conditional execution in the call chain for `WasiStateBuilder`. I tried containing the if statements to the preopen-closure, but the closure is practically required to give a valid path to the `PreopenDirBuilder`.

Forsinge: This commit also deals with second screenshot. If read, write and create all are false, it will not try to preopen a directory.

adelhult (approved):

Issue #159, opened 2023-03-15
Bad styling for html __error modules


adelhult:
**Describe the bug**
The style of the  `__error` module in the html package should be fixed so it does not break when spanning multiple lines. See image:

![image](https://user-images.githubusercontent.com/11508459/225268678-46b15d7a-18f7-4c87-92fa-658a79aa59d7.png)

PR #160, opened 2023-03-15
Fix styling on the html  __error module

adelhult:
Resolves GH-159.


CMDJojo:
Hmm this doesn't seem to be rebased with main...

adelhult:
fixed!

Issue #162, opened 2023-03-15
Add [newpage] and [newline] modules


adelhult:
**Is your feature request related to a problem? Please describe.**
The layout package should support modules for creating line and page breaks.

PR #163, opened 2023-03-15
Add modules for producing line and page breaks

adelhult:
Resolves GH-162. Adds the `[newline]` and `[newpage]` modules in the layout package, I'm open to other name suggestions.

### DIFF
--- a/packages/layout/src/main.rs
+++ b/packages/layout/src/main.rs
@@ -7,6 +7,7 @@ use serde_json::{json, Value};
 fn main() {
     let args: Vec<String> = env::args().skip(1).collect();
     let action = &args[0];
+
     match action.as_str() {
         "manifest" => manifest(),
         "transform" => transform(&args[1], &args[2]),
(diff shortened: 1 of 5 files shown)


packages/layout/src/main.rs
```diff
@@ -70,15 +81,54 @@ fn manifest() {
...
             eprintln!("Package does not support {other}");
         }
     }
 }
 
-fn transform_flex(from: &str, to: &str) {
+fn transform_newline(to: &str, input: Value) {
+    if let Value::String(data) = &input["data"] {
+        if !data.is_empty() {
+            eprintln!("The newline module was fed with some input. Maybe this was a misstake? Consider adding delimiters like this: [newline]().");
```
Forsinge, line 103: Should probably be spelt "mistake"

Forsinge (approved):
Looks good!

This reminds me I forgot to add tests for row and center modules, I will open a new issue for this.

Commented on a rather insignificant typo, up to you if you want to fix it.

packages/layout/src/main.rs
adelhult: Haha, I'll fix the *mistake*!

Forsinge: Good one :)

Issue #153, opened 2023-03-11
Make a good looking HTML document template


adelhult:
**Is your feature request related to a problem? Please describe.**
The HTML package adds support for the `__document` transform. However, it produces a very simple template with no styling at all. We should provide a sensible default look.

**Describe the solution you'd like**
I would suggest to take inspiration from  the article [Some simple ways to make content look good](https://set.studio/some-simple-ways-to-make-content-look-good/) and perhaps the [LaTeX CSS stylesheet](https://latex.vercel.app/).

A bit unrelated. But, to avoid breaking the preview in the web playground we would also want to create a seperate eval function that evaluates a document but without the Document Element to avoid adding a `<body>`-tag and all other stuff that should not be in the preview.

Issue #170, opened 2023-03-20
Support for choosing styling in html package


axoxelol1:
Once the config package is done and there is a way to set variables the package should allow users to choose between available stylesheets. As of writing two are available but it just defaults to `html.css`.

PR #171, opened 2023-03-20
GH-153: Add default html styling

axoxelol1:
This PR  adds two CSS files for the html package, currently it just defaults to the one called `html.css`. Once variables and config is available #170 should be implemented. Also created a eval_no_document function in core that is now used by the playground to avoid getting html body and style tags.

- Resolves: GH-153.

### DIFF
--- a/packages/html/src/main.rs
+++ b/packages/html/src/main.rs
@@ -67,12 +67,11 @@ fn transform_document(doc: Value) -> String {
     let mut result = String::new();
     result.push('[');
 
-    write!(
-        result,
-        "{},",
-        raw!("<html><head><title>Document</title></head><body>")
-    )
-    .unwrap();
+    write!(result, "{},", raw!("<html><head><title>Document</title>")).unwrap();
+
+    write!(result, "{},", raw!("<style>")).unwrap();
+    write!(result, "{},", raw!(include_str!("templates/html.css"))).unwrap();
+    write!(result, "{},", raw!("</style></head><body>")).unwrap();
 
     if let Value::Array(children) = &doc["children"] {
         for child in children {
--- a/packages/html/src/main.rs
+++ b/packages/html/src/main.rs
@@ -67,6 +67,7 @@ fn transform_document(doc: Value) -> String {
     let mut result = String::new();
     result.push('[');
 
+    write!(result, "{},", raw!("<!DOCTYPE html>")).unwrap();
     write!(result, "{},", raw!("<html><head><title>Document</title>")).unwrap();
 
     write!(result, "{},", raw!("<style>")).unwrap();
--- a/packages/html/src/templates/html.css
+++ b/packages/html/src/templates/html.css
@@ -12,7 +12,6 @@
 h1,
 h2,
 h3 {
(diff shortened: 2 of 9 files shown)


packages/html/src/main.rs
```diff
@@ -67,12 +67,11 @@ fn transform_document(doc: Value) -> String {
     let mut result = String::new();
     result.push('[');
 
-    write!(
-        result,
-        "{},",
-        raw!("<html><head><title>Document</title></head><body>")
-    )
-    .unwrap();
+    write!(result, "{},", raw!("<html><head><title>Document</title>")).unwrap();
```
adelhult, line 70: We should add a  `<!DOCTYPE html>` at the very start too!

packages/html/src/templates/html.css
```diff
@@ -0,0 +1,179 @@
...
+  --size-step-1: clamp(1.25rem, calc(1.16rem + 0.43vw), 1.5rem);
+  --size-step-2: clamp(1.56rem, calc(1.41rem + 0.76vw), 2rem);
+  --size-step-3: clamp(1.95rem, calc(1.71rem + 1.24vw), 2.66rem);
+  --size-step-4: clamp(2.44rem, calc(2.05rem + 1.93vw), 3.55rem);
+}
+
+h1,
+h2,
+h3 {
+  font-family: Inter, sans-serif;
```
adelhult, line 15: Users will likely not have Inter installed by default. Either we can opt for a list of more standard fonts or import Inter from [Google fonts](https://fonts.google.com/specimen/Inter?query=inter).

Importing can be done either using a `<link>` tag in the HTML doc or `@import` in your css.
Something like this (note that we likely need to include more font weights then this example) :
```
@import url('https://fonts.googleapis.com/css2?family=Inter:wght@400;700&display=swap');
```

packages/html/src/templates/html.css
```diff
@@ -0,0 +1,179 @@
+:root {
```
adelhult, line 1: Might be worth checking out of we would like to change some settings when printing a html document using a media query:
```css
@media print {
  ...
}
But, I don't have any particular changes in mind!
```

playground/web_bindings/src/lib.rs
```diff
@@ -55,7 +55,7 @@ struct Transpile {
 pub fn transpile(source: &str, format: &str) -> Result<String, PlaygroundError> {
     let result = CONTEXT.with(|ctx| {
         let mut ctx = ctx.borrow_mut();
-        eval(source, &mut ctx, &OutputFormat::new(format))
+        eval_no_document(source, &mut ctx, &OutputFormat::new(format))
```
adelhult, line 58: Transpile is used for multiple options in the playground (Html rendered, Html iframe, LaTeX, Other output formats) but we would only want to remove the document when viewing "Html rendered" so I would suggest creating a new `transpile_no_doc` function or something along those lines

axoxelol1: Good catch, thanks. Will fix

adelhult (approved):

Issue #168, opened 2023-03-19
CLI support live preview


adelhult:
We should add a live preview feature to the CLI which would allow you to only specify a input file but not a output file.

Running:
```sh
modmark test.mdm
```
Would assume the output to be html and serve a web based preview and use web sockets to auto refresh the page (much like how the npm package live-server) works.

PR #172, opened 2023-03-20
Adelhult/gh 168/0

adelhult:
Resolves GH-168. Might of ended up a bit over-kill but as a week-end project I tried to implement something along the lines of the `live-server` utility directly into the ModMark cli.

You can now use the command-line tool without providing a output file `modmark myfile.md` and a live updating preview will be served.  This CLI prints the address of a website where the document can be previewed. Changes made in the document will be updated without having to refresh the browser window. (I use Websockets to send messages to the browser about the new changes.)

This new system will also later allow us to send information about warnings and errors and inject them into the preview for a nicer writing experience were you don't have to keep an eye on the terminal.

I'm sure there are many bugs since I've never used tokio or warp before, but feel free to take a look and leave comments. 

[Screencast from 2023-03-20 15-20-23.webm](https://user-images.githubusercontent.com/11508459/226369345-7cc85795-30fa-4911-b565-fea91cb91e4b.webm)


CMDJojo:
Haha that is really cool! Does it have the same behaviour as doing `modmark --watch in.mdm out.html & live-server out.html`?

cli/src/main.rs
```diff
@@ -37,172 +54,393 @@ struct Args {
...
+    print_result(&compilation_result, &args)?;
+
+    Ok(())
+}
+
+/// Choose a free port for hosting the html live preview
+fn get_port() -> Result<u16, CliError> {
+    PREVIEW_PORT
+        .get_or_init(|| portpicker::pick_unused_port())
+        .ok_or(CliError::NoFreePorts)
```
CMDJojo, line 277: It is nice to have a check if there are free ports available and fail the live server otherwise, however, assuming that there are no free ports (or that we block modmark from creating sockets, which wouldn't be too unreasonable for some users depending on system settings), we can no longer compile to html with the watch flag at all; `modmark --watch in.mdm out.html` exits immediately saying "No free ports" without compiling the document

cli/src/main.rs
```diff
@@ -37,172 +54,393 @@ struct Args {
...
+    save_result(&compilation_result, &args)?;
+    print_result(&compilation_result, &args)?;
+
+    Ok(())
+}
+
+/// Choose a free port for hosting the html live preview
+fn get_port() -> Result<u16, CliError> {
+    PREVIEW_PORT
+        .get_or_init(|| portpicker::pick_unused_port())
```
CMDJojo, line 276: You can use a function handle directly
```suggestion
        .get_or_init(portpicker::pick_unused_port)
```

cli/src/main.rs
```diff
@@ -37,172 +54,393 @@ struct Args {
...
+    document: PreviewDoc,
+    connections: PreviewConnections,
+    current_path: PathBuf,
+) -> impl Filter<Extract = (impl Reply,), Error = Rejection> + Clone + 'static {
+    let connections = warp::any().map(move || connections.clone());
+    let document = warp::any().map(move || document.clone());
+
+    let websocket = warp::path("ws").and(warp::ws()).and(connections).map(
+        |ws: warp::ws::Ws, connections: PreviewConnections| {
+            ws.on_upgrade(move |socket| handle_connection(socket, connections.clone()))
```
CMDJojo, line 290: I think this clone is redundant since you explicitly `move` the variables in to the closure
```suggestion
            ws.on_upgrade(move |socket| handle_connection(socket, connections))
```

cli/src/main.rs
```diff
@@ -37,172 +54,393 @@ struct Args {
...
+        },
+    );
+
+    let preview = warp::path::end().and(document).map(|document: PreviewDoc| {
+        // Inject a JS script after the end of the </body> tg
+        let html = document.lock().unwrap();
+        let modified_html = RE_INJECTION.replace(
```
CMDJojo, line 297: This works fine for the special case with HTML, but previewing a latex document won't inject any update code and thus won't do live updates, even though the terminal output says it does.

CMDJojo: Also, I am unsure why you are using regex in this case. While it might be nice for more advanced matches further down the line, you are replacing "</body>" captured by a regex with the literal "</body>" concatenated with your injection code. Since the second part is still a literal, I think we can do the search-and-replace using a literal as well, that would probably be more performant and won't require regex. As an alternative, consider replacing with the capture group 0 concatenated with your injection.

cli/src/main.rs
```diff
@@ -37,172 +54,393 @@ struct Args {
...
+            &html,
+            concat!("</body>\n", include_str!("./preview_injection.html")),
+        );
+        warp::reply::html(modified_html.to_string())
+    });
+
+    let working_directory = warp::fs::dir(current_path);
+
+    let routes = working_directory.or(websocket).or(preview);
```
CMDJojo, line 306: You don't have to bind this to a variable, you can implicitly return it without binding

cli/src/main.rs
```diff
@@ -37,172 +54,393 @@ struct Args {
...
+
+    routes
+}
+
+async fn handle_connection(socket: WebSocket, connections: PreviewConnections) {
+    // Transmitter and reciever for the websocket
```
CMDJojo, line 312: Typo
```suggestion
    // Transmitter and receiver for the websocket
```

cli/src/main.rs
```diff
@@ -37,172 +54,393 @@ struct Args {
...
+    let (mut ws_tx, mut ws_rx) = socket.split();
+
+    // Transmitter and receiver for communication with the rest of the program
+    let (tx, rx) = mpsc::unbounded_channel();
+    let mut rx = UnboundedReceiverStream::new(rx);
+
+    // Get a id and add this connection to HashMap of all connections
+    let id = CONNECTION_ID_COUNTER.fetch_add(1, Ordering::Relaxed);
```
CMDJojo, line 320: Seems like Aquire is more appropriate, so we get fencing
```suggestion
    let id = CONNECTION_ID_COUNTER.fetch_add(1, Ordering::Aquire);
```

cli/src/main.rs
```diff
@@ -37,172 +54,393 @@ struct Args {
...
+    if args.use_html_preview() {
+        let port = get_port()?;
+        println!("Live preview available at: http://localhost:{port}");
+    }
 
-        println!(
-            "Your file can be found at {}\nSave your file to recompile changes.",
-            location.display()
-        );
+    println!("");
```
CMDJojo, line 162:
```suggestion
    println!();
```

cli/src/main.rs
```diff
@@ -37,172 +54,393 @@ struct Args {
...
+            Ok(event) => {
+                // If we also generate a output file, discard any changes from that file
+                if let Some(output_path) = ABSOLUTE_OUTPUT_PATH.get() {
+                    if event.paths.contains(output_path) {
+                        continue;
+                    }
+                }
+
+                // We only care about changes from when files are created, removed or modified
+                if let EventKind::Create(_) | EventKind::Modify(_) | EventKind::Remove(_) =
```
CMDJojo, line 417: Maybe using the methods for checking kind on event.kind directly looks better?
```rs
if event.kind.is_create() || event.kind.is_modify() || event.kind.is_remove() {...}
```

CMDJojo (changes requested):
Good work! Nice to have an live preview, it looks good! I haven't fully checked all the code yet, but I have commented on some small things. Here are some additional, overall things I noticed when playing around initially:

 * You can do `modmark in.mdm -f latex` to get the result in latex, however, it is not a live preview as the output states. It would probably be best to make a simple html wrapper around it, or something, but I think it is bad to state that it is a live preview when it isn't.
 * Bringing in `warp` gives a future incompatibility warning because of its dependency on `buf_redux` and `multipart`. Check if there is an better alternative and, if not, check if there are patches available that we can apply

Issue #173, opened 2023-03-20
Improve start-up time by caching pre-compiled WASM modules


adelhult:
A big part of the start-up time is due to us needing to compile the entire prelude of standard packages into something that we actually can instantiate and run. Wasmer has support for [serializing](https://docs.rs/wasmer/latest/wasmer/struct.Module.html#method.serialize) pre-compiled wasm-modules, if we cache these we would likely improve the start-up time by a lot. 

We should also investigate this crate further: https://crates.io/crates/wasmer-cache

PR #174, opened 2023-03-22
GH-173: Add `precompile_wasm` feature

CMDJojo:
This PR adds the feature `precompile_wasm` to `core`, which may be used when targeting `native` and using `bundle_std_packages`. When enabled, as a part of the build script which builds the packages, another step is added which runs the packages though the Cranelift compiler (which is the one we use in Wasmer), and saves that Cranelift-compiled module instead of the actual WebAssembly file. This file, that has already been compiled, is referred to by me as "precompiled" and is using the `.wir` file format (I made that up, Wasmer Internal Representation). The main benefit this has is improving start-up time for the ModMark binary, since Wasmer doesn't have to compile the WebAssembly modules.

Due to technical limitations in Cargo and to developer wrongdoings in Wasmer, the actual implementation of the compilation step is done using `wasmer-compiler` and `wasmer-compiler-cranelift` rather than just using `wasmer`. More details are in `Cargo.toml` and `build.rs`.

With this, there are three options for distributing a binary; without bundled packages (standalone), with bundled packages, and with precompiled packages. Here are how they compare (only basic testing has been done, compiled with `--release`, running `modmark in.mdm out.html`):
| Type | File size | Time to compile `example.mdm` |
|------|---------|----------------------------------|
| Standalone | 9.9MB | n/a |
| Bundled | 29.9MB | 2.05 s |
| Precompiled | 36.8MB | 0.02 s |

Since this gives such a big time improvement, and doesn't add much to the compile time (it compiles the wasm files to wir files once, which Cranelift would have to do on startup anyways), this is enabled by default (but just like bundle_std_packages it can be disabled by `--no-default-features`). Ready for review immediately!

Resolves GH-173

### DIFF
--- a/core/build.rs
+++ b/core/build.rs
@@ -89,13 +89,13 @@ fn precompile_wasm(name: &str, output_path: &Path, engine: &Engine) {
         .join(name)
         .join("wasm32-wasi")
         .join("release")
-        .join(name.to_string() + ".wasm");
+        .join(format!("{name}.wasm"));
 
     let out_path = output_path
         .join(name)
         .join("wasm32-wasi")
         .join("release")
-        .join(name.to_string() + "-precompiled.wir");
+        .join(format!("{name}-precompiled.wir"));
 
     // Much of this code is taken from Module::serialize and by following what it does, function
     // calls etc
--- a/core/build.rs
+++ b/core/build.rs
@@ -5,6 +5,19 @@ use std::{
     process::Command,
 };
 
+#[cfg(all(
+    feature = "bundle_std_packages",
+    feature = "native",
+    feature = "precompile_wasm"
+))]
+use wasmer_compiler::{ArtifactCreate, Engine, EngineBuilder};
+#[cfg(all(
+    feature = "bundle_std_packages",
+    feature = "native",
+    feature = "precompile_wasm"
+))]
+use wasmer_compiler_cranelift::Cranelift;
+
 fn main() {
     // build packages if we want to bundle them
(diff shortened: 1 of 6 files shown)


core/build.rs
```diff
@@ -45,10 +66,47 @@ fn build_packages() {
...
+    feature = "precompile_wasm"
+))]
+/// This function pre-compiles a wasm file for a package with the given name, and makes a new file
+/// `{name}-precompiled.wasm` which holds the serialized data for the precompiled package.
+fn precompile_wasm(name: &str, output_path: &Path, engine: &Engine) {
+    let in_path = output_path
+        .join(name)
+        .join("wasm32-wasi")
+        .join("release")
+        .join(name.to_string() + ".wasm");
```
adelhult, line 92: Might be cheaper/cleaner to do `&format!("{name}.wasm")` but does not really matter. Or perhaps you can push and then use set_extension to the pathbuf.

adelhult (approved):
Great work! Really nice to see the improved start-up speed. I think the code it ended up quite nice regardless of the trouble with wasmer feature flags. Left a minor nit-pick that you can ignore :sweat_smile: , I'm not even sure it makes a difference at all.

core/build.rs
CMDJojo: Yeah, I can do format strings, sure! That's what I usually do, however I thought that since this only was one extra call to .to_string(), it wouldn't make that much of a difference. But I fixed it here and at the other place

PR #176, opened 2023-03-22
GH-115: Add latex support for links

hugomardbrink:
- Resolves: GH-115.


CMDJojo (commented):
I think the `\href` requires the href package, so the output latex doesn't really compile as-is. Maybe add that to the prelude (like we do with `\usepackage{ulem}`)

CMDJojo (approved):
LGTM!

Issue #183, opened 2023-03-23
Relax restrictions for module argument values


CMDJojo:
**Is your feature request related to a problem? Please describe.**
As a part of #180 which was solved by #181, the optional arguments `caption` and `label` was added to the `[table]` module. The `label` is intended to be used for `\ref`-like references used in LaTeX (supported by anchor tags in HTML), so that you could do something like:
```
[table label="table:abc"]
1|2|3
4|5|6

See table [ref] table:abc
```
This works fine in the current state, but we have a restriction on module argument values where you may only use alphanumeric characters if you don't encase the content in quotes. This was done for performance reasons and for having to explicitly wrap something including a comma, since we before-hand supported both comma-separated and space-separated arguments. This will make LaTeX-style labels, like `table:name`, somewhat cumbersome to use. Additionally, it is actually shorter to write `[table "" table:abc]` (to have an empty string as the first argument, which is supposed to be the caption), but that is not allowed now.

**Describe the solution you'd like**
I suggest we change the restrictions of module argument values to be able to include any non-whitespace characters but `=` (so `a=b=c` isn't allowed to avoid confusion) and `]` (which is needed to parse `[table abc]` without treating the closing `]` as a part of the argument), and that we allow empty strings in quotation marks. This will still treat multiple spaces like `[table     ""     ""]` as one separator.

PR #184, opened 2023-03-23
GH-183: Relax restrictions for argument value parsers

CMDJojo:
Resolves GH-183. This PR also includes tests for these new corner-cases


CMDJojo:
Some examples of module expressions that works after this change that didn't before, that you can test in the playground, is:
```
[table caption=😀]
1|2|3
4|5|6


[table 😀]
1|2|3
4|5|6


[table caption=""]
1|2|3
4|5|6


[table ""]
1|2|3
4|5|6
```

adelhult (approved):

Issue #127, opened 2023-03-02
Make cli support different registry links


hugomardbrink:
**Is your feature request related to a problem? Please describe.**
If the user want to use another registry this should be supported in the cli

PR #185, opened 2023-03-23
GH-127: Add flag to choose a different registry

CMDJojo:
Resolves GH-127. This PR adds a `--registry` flag (shorthand `-r`) which lets the user specify an URL to a different registry that the user wants to use rather than the default registry we provide. This is achieved by adding a `registry: String` field to the PackageManager which holds the registry URL, and constructing the `Context` with a `PackageManager` created after the arguments has been parsed. It changes the `Lazy<...<Context<PackageManager>>>` for `OnceCell<...<Context<PackageManager>>>` since we can't just create a `Context` immediately; we need to parse the arguments before creating one. At all places where the `Context` is used, it is forcefully unwrapped since it is guaranteed to be set immediately after parsing the arguments (or, if it fails to create a `Context`, it will `panic!` as it did previously).

Additionally, this PR includes some small improvements to the PackageManager, such as using the `json` feature in `Reqwest` to include `serde_json` and being able to call `response.json()` instead of `response.text()` and parsing the json separately afterwards. It also bumps the version of `directories` to `5.0.0`.

### DIFF
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -8,7 +8,7 @@ use crossterm::{
     terminal, ExecutableCommand,
 };
 use notify::{Config, Event, PollWatcher, RecommendedWatcher, RecursiveMode, Watcher, WatcherKind};
-use once_cell::sync::Lazy;
+use once_cell::sync::OnceCell;
 
 use error::CliError;
 use modmark_core::{context::CompilationState, OutputFormat};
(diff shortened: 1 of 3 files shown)


cli/src/main.rs
```diff
@@ -32,6 +32,9 @@ struct Args {
     #[arg(short = 'f', long = "format", help = "The output format of the file")]
     format: Option<String>,
 
+    #[arg(short = 'r', long = "registry", help = "An URL to the registry to use")]
```
adelhult, line 35: Typo (consonant sounding "U")
```suggestion
    #[arg(short = 'r', long = "registry", help = "A URL to the registry to use")]
```

adelhult (approved):

cli/src/main.rs
CMDJojo: Fixed!

Issue #118, opened 2023-03-01
Delegate __math tags to math module


CMDJojo:
**Is your feature request related to a problem? Please describe.**
The parser sees `$$` as a math tag. This means that `$$x^2$$` will be `Tag [ Text("x^2") ]` (in pseudo-structure). Currently, neither the HTML language package in `main` nor the LaTeX language package as a part of #108 handles this. As a part of #117, a math module is (expected to be) added, which correctly parses LaTeX math such as `[math]$x^2 + y^2 = z^2$` and outputs both HTML and LaTeX.

**Describe the solution you'd like**
It would be nice if the `__math` tags (`$$`) gets converted to inline `[math]` modules, so it delegates the actual presentation to the `math` package. Since that works fine, as long as the tag only contains text, it would be feasibly and probably preferable to delegate to there.

**Describe alternatives you've considered**
We could do alternative implementations for the `__math` tag, possibly in the same way as the `math` package handles it. It would be an unnecessary copy-paste of the code and may, if implemented differently, result in different outputs which is unexpected.

PR #189, opened 2023-03-23
GH-118: Delegate text-only math tags to math module

CMDJojo:
This PR resolves GH-118 by adding support for math tags `$$` containing text only, by taking the text and making a `math` module containing the very same text. More detailed explanation on why to delegate it to the `math` module rather than implementing support for it in the language packages directly can be found in the [original issue](https://github.com/modmark-org/modmark/issues/118). The support is added to both the HTML and LaTeX language packages. Even though it could in theory be implemented in the `math` package, I think it makes more sense to add them to the language packages which should be responsible for handling all tags.

All text from within the math tags gets concatenated, and that is the input to the `math` module. Since the tags are non-recursive, only text and modules may occur within, and if a module appears within an warning is generated. I think this is ready for review as-is! It also includes 6 tests, testing "normal operation", math tags with modules inside, and empty math tags, for both of the languages.

### DIFF
--- a/packages/html/src/main.rs
+++ b/packages/html/src/main.rs
@@ -55,6 +55,7 @@ fn transform(from: &str) -> String {
         "__underlined" => transform_tag(input, "u"),
         "__strikethrough" => transform_tag(input, "del"),
         "__paragraph" => transform_tag(input, "p"),
+        "__math" => transform_math(input),
         "__document" => transform_document(input),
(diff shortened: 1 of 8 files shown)


adelhult (approved):
Approved! I think we will need to revisit this once the work on the scheduler has started since this code works on the assumption that the children of a parent node are not evaluated yet.

Issue #193, opened 2023-03-27
Improve styling of playground package view


adelhult:
**Describe the bug**
The styling in the package view is not the best and could enjoy a face-lift.
![image](https://user-images.githubusercontent.com/11508459/228025787-4529cfb9-54ea-4a80-abb7-c94620474a4f.png)

PR #194, opened 2023-03-27
Improve package view styling

adelhult:
Resolves GH-193 with a splash of color for the package view.

<table>
<tr>
<td>Before</td>
<td>After</td>
</tr>
<tr>
<td>

</td>

<td>

</td>
</tr>

### DIFF
--- a/playground/static/main.js
+++ b/playground/static/main.js
@@ -24,7 +24,7 @@ async function compilerAction(action) {
         compiler_callback = res;
         compiler_failure = rej;
     });
-    compiler.postMessage({seq: ++seq, ...action});
+    compiler.postMessage({ seq: ++seq, ...action });
     return promise;
 }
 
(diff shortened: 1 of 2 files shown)


CMDJojo:
I think that is a very nice improvement, I especially love the colors! One small comment is that I think it sounds better saying "supports all formats" instead of "supports all" - you can't really tell what "supports" refers to if you haven't seen other examples saying "supports html". But good job!

playground/static/main.js
```diff
@@ -158,32 +158,56 @@ function toggleView() {
...
+            color = "#8B85C1";
+        } else if (type == "Float") {
+            color = "#C3423F";
+        }
+
+        return `<span style="background-color: ${color};" class="type">${type_name}</span>`
+    }
+
+    const escape_default = (type, default_value) => {
+        if (Array.isArray(type) || type == "String") {
```
CMDJojo, line 183: Suggestion: Disable type coercion
```suggestion
        if (Array.isArray(type) || type === "String") {
```

playground/static/main.js
```diff
@@ -158,32 +158,56 @@ function toggleView() {
...
 
-    const type_str = (type) => {
+    const type_annotation = (type) => {
+        let type_name = type;
+        let color = "";
         if (Array.isArray(type)) {
-            return type.join("/");
+            type_name = type.join("/");
+            color = "#90A959";
+        } else if (type == "String") {
```
CMDJojo, line 169: Suggestion: Disable type coercion
```suggestion
        } else if (type === "String") {
```

playground/static/main.js
```diff
@@ -158,32 +158,56 @@ function toggleView() {
...
+            color = "#2A6041";
+        } else if (type == "Unsigned integer") {
```
CMDJojo, line 171: Suggestion: Disable type coercion
```suggestion
        } else if (type === "Unsigned integer") {
```

CMDJojo (commented):
Nice job with this PR! It looks much better now!
I have left some comments on parts that may be improved, but overall, really good job! Out of curiosity, what does the `#package-content .expanded .details { max-height: 2000px; }` do and what is it for?

CMDJojo (approved):
LGTM

PR #199, opened 2023-03-29
GH-198: Ignore runtime errors where exit code was 0

### LINKED ISSUE #198: Packages exiting early shows incorrect error
When a package wants to give an error it is supposed to not print anything to stdout and only print the error to stderr. This would be a lot easier if the package can exit with a 0 code but without finishing the function. For example
```
eprintln!("Oh noes something went wrong");
std::process:exit(0);
```
Doing this will give an error that looks like:
image

Instead it should show the error message that was printed to stderr by the package.

axoxelol1:
RuntimeError did not seem to be typed in any meaningful way so unfortunately I think the best way to do this might just be a string comparison.

Resolves #198

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -602,13 +602,17 @@ impl<T> Context<T> {
         let fn_res = main_fn.call(&mut store, &[]);
 
         if let Err(e) = fn_res {
-            // An error occurred when executing Wasm module =>
-            // it probably crashed, so just insert an error node
-            return Ok(Left(create_issue(
-                true,
-                format!("Wasm module crash: {e}"),
-                &input_data,
-            )));
+            // TODO: See if this can be done without string comparison
+            let error_msg = e.to_string();
+            if !error_msg.contains("WASI exited with code: 0") {
+                // An error occurred when executing Wasm module =>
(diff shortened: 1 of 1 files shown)


Forsinge:
This seems relevant: https://github.com/wasmerio/wasmer/blob/master/examples/early_exit.rs

If anything it's a more "proper" way to handle it, though I can't get it working myself for some reason. Food for thought at least!

In any case I was trying to apply it like this:

```rust
match e.downcast::<ExitCode>() {
    Ok(ExitCode(0)) => {}
    Ok(exit_code) => {...}
    Err(e) => {...} // a failed downcast returns the original error
}
```

Forsinge (approved):
For our purposes I think string comparison is fine, unless you want to try to get the other method working. Maybe add a TODO if not

CMDJojo:
I think it looks good as well! The only thing I may add is to maybe do the string conversion just once and store it in a local variable, but otherwise it seems fine!

axoxelol1:
@CMDJojo I changed it, looks good?

CMDJojo:
Looks good! It was just a small change, I doubt the previous version would be expensive anyways, but looks good! Ready to merge

Issue #201, opened 2023-03-30
Filesystem functions read_file and create_file panic when given invalid paths


Forsinge:
**Describe the bug**
 Both of these functions were written under the assumption that they would only be used for downloading files with valid paths in our playground. As it turns out they have other use cases, where the provided paths have no such guarantees.

PR #202, opened 2023-03-30
GH-201: Avoid unwraps in read_file and create_file

Forsinge:
Resolves: #201


CMDJojo (approved):
LGTM. My favorite is line 156, what's yours? Leave your answers in the comment section down below

Issue #206, opened 2023-04-04
Improve HTML document template


adelhult:
The HTML document templates contains a few bugs related to spacing between elements and margins due to a missing article element amongst other things.

PR #207, opened 2023-04-04
Improve HTML document template

adelhult:
Resolve GH-206. Fixes the text margin for smaller screens and print. I also added the Lora font and imported it from Google Fonts for a consistent style across browsers.


axoxelol1 (approved):
Looks good =)

PR #210, opened 2023-04-04
Improve files package

### LINKED ISSUE #179: Improve files package
The files packages included in #158 was primarily made as proof-of-concept.  A number of improvements can be made to the modules inside, such as adding arguments for customization and avoiding embedding base64 in the HTML for images.

### LINKED ISSUE #209: Add LaTeX support in files package
Part of #179 

This issue focuses on adding transforms from [textfile], [image] and [include] to LaTeX, primarily to enable report-writing with LaTeX as output. Generalizations and extensions to these transforms should be included in #179

Forsinge:
Resolves #209, resolves #179 

This PR adds transforms to LaTeX for all modules inside the files package.

The [image] module also includes `caption`, `label` and `width` as arguments. I'm a little unsure to optimally handle the user not specifying `width`, and what it should default to in that case. Currently it defaults to "width=1.0\textwidth".

### DIFF
--- a/packages/files/src/main.rs
+++ b/packages/files/src/main.rs
@@ -1,8 +1,10 @@
 use base64::{engine::general_purpose, Engine as _};
 use serde_json::{json, Value};
 use std::env;
+use std::ffi::OsStr;
 use std::fs;
 use std::io::{self, Read};
+use std::path::Path;
 
 fn main() {
     let args: Vec<String> = env::args().skip(1).collect();
@@ -23,7 +25,7 @@ fn manifest() {
             {
             "name": "files",
             "version": "0.1",
-            "description": "This package provides file access",
+            "description": "This package provides file access.",
             "transforms": [
                 {
                     "from": "textfile",
@@ -37,18 +39,12 @@ fn manifest() {
                         {
                             "name": "caption",
                             "default": "",
-                            "description": "The caption for the image"
+                            "description": "The caption for the image."
                         },
                         {
                             "name": "label",
                             "default": "",
-                            "description": "The label to use for the image, to be able to refer to it from the document"
-                        },
-                        {
-                            "name": "type",
-                            "type": ["image", "svg"],
-                            "default": "image",
-                            "description": "The type of source file. SVG is currently only supported in LaTeX output."
+                            "description": "The label to use for the image, to be able to refer to it from the document."
                         },
                         {
                             "name": "width",
@@ -60,6 +56,12 @@ fn manifest() {
                                 For LaTeX this is ratio to the document's text area width. \
                                 For HTML this is ratio to the width of the surrounding figure tag (created automatically).\
                                 "
+                        },
+                        {
+                            "name": "embed",
+                            "default": "false",
+                            "type": ["true", "false"],
+                            "description": "Decides if the provided image should be embedded in the HTML document."
                         },
                     ],
                 },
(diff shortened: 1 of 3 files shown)


axoxelol1:
Maybe not in the scope of this PR but it would be nice if the html transform has support for the same arguments. I think we should try to avoid having different arguments for different output formats if possible. I will try to review this PR properly tomorrow if nobody else has gotten to it =)

CMDJojo:
@forsinge nice PR, it is nice to have LaTeX support for files! One thing I would like to see changed is the argument order for `image`; having the first argument be "caption" and the second "label" will make it consistent with the table module, which I think would be good. Additionally, we may consider to allow labels and captions for HTML output as well, "caption" can be implemented with <figcaption>, see https://www.w3schools.com/TAGS/tag_figcaption.asp, and "label" can be implemented by an anchor tag, <a>. It is not a strict requirement but it is nice to be able to have the same code compile first to HTML for a preview and then being able to change it to LaTeX without needing to change very much in the Modmark input.

Forsinge:
Fair points, I'll start working on it now. I'm not entirely sure how you mean I should use the anchor tag. Isn't the anchor tag used elsewhere in the HTML document, referring to the `id=label` i set in ``?

Forsinge:
The `[image]` -> HTML transform now has the same arguments as LaTeX. SVG is still unsupported for HTML, because it'll likely require editing the SVG itself to scale it (if you have other ideas, LMK!). Eventually, as a part of #179 I want to implement a separate `[svg]` module, but I feel like the changes are growing out of scope for this PR.

packages/files/src/main.rs
```diff
@@ -27,17 +27,45 @@ fn manifest() {
...
                 {
                     "from": "image",
-                    "to": ["html"],
-                    "arguments": [],
+                    "to": ["html", "latex"],
+                    "arguments": [
+                        {
+                            "name": "caption",
+                            "default": "",
+                            "description": "The caption for the image"
```
adelhult, line 40:
```suggestion
                            "description": "The caption for the image."
```

packages/files/src/main.rs
```diff
@@ -27,17 +27,45 @@ fn manifest() {
...
+                        },
+                        {
+                            "name": "label",
+                            "default": "",
+                            "description": "The label to use for the image, to be able to refer to it from the document"
```
adelhult, line 45:
```suggestion
                            "description": "The label to use for the image, to be able to refer to it from the document."
```

packages/files/src/main.rs
```diff
@@ -27,17 +27,45 @@ fn manifest() {
...
+                        },
+                        {
+                            "name": "type",
+                            "type": ["image", "svg"],
+                            "default": "image",
```
adelhult, line 50: I think we should infer the type based on the file extension, but we can leave that for later and open another issue

packages/files/src/main.rs
```diff
@@ -93,8 +125,31 @@ fn transform_image(input: Value, to: &str) {
...
+                    };
+
+                    let percentage = (width * 100.0).round() as i32;
+                    let style = format!("style=\"width:{percentage}%\"");
+
+                    let mut v = vec![];
+
+                    v.push(String::from("<figure>"));
+                    v.push(format!(
+                        "<img src=\"data:image/png;base64,{encoded}\" {id} {style}/>"
```
adelhult, line 145: Maybe we should default to to just do `` instead of embedding it? But we can deal with this later too!

packages/files/src/main.rs
```diff
@@ -93,8 +125,31 @@ fn transform_image(input: Value, to: &str) {
...
+                    } else {
+                        format!("id=\"{label}\"")
+                    };
+
+                    let percentage = (width * 100.0).round() as i32;
+                    let style = format!("style=\"width:{percentage}%\"");
+
+                    let mut v = vec![];
+
+                    v.push(String::from("<figure>"));
```
adelhult, line 143: Maybe it's a nice idea to only include the figure element if a caption is provided?

packages/files/src/main.rs
```diff
@@ -93,8 +125,31 @@ fn transform_image(input: Value, to: &str) {
             match fs::read(path) {
                 Ok(contents) => {
                     let encoded: String = general_purpose::STANDARD_NO_PAD.encode(contents);
-                    let html = format!("<img src=\"data:image/png;base64, {encoded} \"/>");
-                    let json = json!({"name": "raw", "data": html}).to_string();
+                    let width = input["arguments"]["width"].as_f64().unwrap();
+                    let caption = input["arguments"]["caption"].as_str().unwrap();
```
adelhult, line 129: I think we should support ModMark markup inside of the caption, so perhaps wrap the caption inside of a `inline_content` module?  (Or alternativly a `__text` module. With the current behaviour we allow users to inject raw content directly, see screenshot.  

image

adelhult (commented):
Looks great! Left a few comments, you can save most of them for another PR if you want, but I think we should fix the bug with the raw caption before we rebase.

packages/files/src/main.rs
Forsinge: Maybe I'm missing something obvious but how would the browser/HTML document find the file? I discussed creating JavaScript blobs with Jonathan a few days ago, but that seems rather messy to me.

adelhult: They don't ; )

Most markup languages just links to file path and leaves the user responsible that those files are present when viewing the transpiled document too. But, I think it' great that we support embedding the images too!

Forsinge: I see. In that case it seems nice to add an embed argument, however that once again runs into the issue of having different arguments for HTML and LaTeX. Maybe that's fine?

adelhult: yeah, it's rather unfortunate but I guess the most important part is ensuring that the document compiles regardless of the output, so if you add a extra argument it should exist for latex too and just be a no-op.

packages/files/src/main.rs
Forsinge: This won't be necessary if we add the `[svg]` module I was talking about (I think.. are there other filetypes worth considering?)

You could also argue that it's nicer to skip the distinction and treat everything as an image, from the user's POV. Which do you think is better?

adelhult: Good point! I don't have a super strong opinion so I think it's your call : )

packages/files/src/main.rs
Forsinge: I'm aware I marked this as resolved earlier, but I've realized the `<figure>` tag slightly resizes the image inside. To me it doesn't seem very intuitive for the user that it gets resized because of a caption. Thoughts?

CMDJojo: I think it is nice to have it being a figure anyways. This solves one problem I found very annoying; if you have an 4k image (or another large image, which you may get from a phone or something) then the whole image would be shown in the source size (which is huge, you can impossibly see the whole image at once but just one corner). Additionally, it was very dependent on screen resolution. The addition of `<figure>` makes it (relative to) the width of the page in both HTML and LaTeX, which I find consistent and good; I think this is more like what an user expects than for one corner of the image to be shown.

Actually, I chimed in to try out this very thing; since I assumed the previous behavior was present, I was curious how the size argument would work, like if you made it `1.0` it would be great for LaTeX but bad for HTML and vice versa if you set it to the likes of `0.2`. But I clearly see that my worries are solved with the figure tag, so I think that is great! The only problem is that it upscales small images, but I think it is fine anyways.

Good job with the module overall!

Forsinge:
Thanks for all the input! I believe I've fixed everything you've commented on. This PR has now more or less grown into what I had in mind for #179 , so unless you have other ideas for that issue I will close that as well.

CMDJojo:
> Thanks for all the input! I believe I've fixed everything you've commented on. This PR has now more or less grown into what I had in mind for #179 , so unless you have other ideas for that issue I will close that as well when this is merged

Great! Will have a look at the code later, expect it done within 2h, but I doubt there are any glaring problems. Feel free to edit the PR message and add "Resolves GH-179" and that issue will be linked to this PR as well

CMDJojo (approved):
Looks great! Thanks for fixing all the previously mentioned comments

Issue #212, opened 2023-04-06
Add [if] module for conditional compilation


adelhult:
We will need a module for conditional compilation in the thesis. It might look something like this:

```
[if format=latex]
[include](latex-stuff.mdm)
```

Side note, for now this will have to be a core module since we don't allow ordinary modules to support all output formats yet.

PR #213, opened 2023-04-06
Add `[if]` module

adelhult:
Resolves GH-212.


CMDJojo:
Hmm, I would argue that it is better to do this as a normal external module, maybe in a package called flow control or something similar?

adelhult:
Close in favour of PR #233.

PR #214, opened 2023-04-09
Add latex support to the code module

bruhng:
Resolves: GH-113
Adds latex support to the code module by implementing a custom code highlighter for latex. The latex code is not preatty but it works.


bruhng:
image

CMDJojo:
Nice work! It will be nice to get LaTeX support for highlighted code, so we get closer to HTML/Latex parity. Something I saw when looking through the code was the changes to the manifest lines 37 and 38, it seems like you changed back the argument type from uint to string, is there any reason for this? I think it makes more sense for font size and tab size to still be ints.

bruhng:
> Nice work! It will be nice to get LaTeX support for highlighted code, so we get closer to HTML/Latex parity. Something I saw when looking through the code was the changes to the manifest lines 37 and 38, it seems like you changed back the argument type from uint to string, is there any reason for this? I think it makes more sense for font size and tab size to still be ints.


That was not my intention, i probably made a mistake while rebasing

i will have to fix it some other day

packages/code/src/main.rs
```diff
@@ -67,25 +67,25 @@ fn transform_code(to: &str) {
...
 
-            let code = input["data"].as_str().unwrap();
-            let lang = get_arg!(input, "lang");
-            let font_size = input["arguments"]["font_size"].as_u64().unwrap();
-            let tab_size = input["arguments"]["tab_size"].as_u64().unwrap();
-            let theme = get_arg!(input, "theme");
-            let bg = get_arg!(input, "bg");
+    let code = input["data"].as_str().unwrap();
+    let lang = get_arg!(input, "lang");
+    let font_size = get_arg!(input, "font_size");
```
CMDJojo, line 78: This would be changed back as a result of reverting lines 37 and 38

CMDJojo (changes requested):
Hi and thank you for your PR!

It is very nice to get closer to feature parity with HTML and LaTeX! Good job! It seems to work great in multiple regards. However, I think there is quite a few mistakes done when rebasing that needs to be fixed, and some other style-related things I think could be changed. I have not loaded the code in an editor and checked it out, rather I just commented on some things I saw at first glance.

But, as said, nice job and good work with fixing the issues I see you mentioned you had before!

Some other things: It doesn't look like your LaTeX code supports the font size argument, is this intentional? Also, it seems like the code isn't in its own block:
```
This is some text

[code lang=py]
def foo():
    print("abc")
```
gives this:
image

packages/code/src/main.rs
```diff
@@ -94,13 +94,113 @@ fn transform_code(to: &str) {
...
+            print!("{}", highlight_latex(code, lang, theme, tab_size));
+        }
         other => {
             eprintln!("Cannot convert code to {other}");
         }
     }
 }
 
-fn get_style(
+fn highlight_latex(code: &str, lang: &String, tm: &str, tab_size: u64) -> String {
```
CMDJojo, line 109: I think the naming may be confusing, `highlight_latex` vs `get_highlighted_html`, it seems to me that they do similar things, right? They may be named in a way which is more clear that they are equivalent.

Also, `&str` is better than `&String`

packages/code/src/main.rs
```diff
@@ -126,7 +226,7 @@ fn get_style(
     style
 }
 
-fn get_highlighted(code: &str, lang: &String, tm: &str) -> (String, Option<Color>) {
+fn get_highlighted_html(code: &str, lang: &String, tm: &str) -> (String, Option<Color>) {
     let ss = SyntaxSet::load_defaults_newlines();
     let ts = ThemeSet::load_defaults();
     let theme = match tm {
```
CMDJojo, line 232: We have two (to me) identical theme matches, and loading of `SyntaxSet` and `ThemeSet`, might be nicer to not do that in two places

packages/code/src/main.rs
```diff
@@ -94,13 +94,113 @@ fn transform_code(to: &str) {
...
             eprintln!("Cannot convert code to {other}");
         }
     }
 }
 
-fn get_style(
+fn highlight_latex(code: &str, lang: &String, tm: &str, tab_size: u64) -> String {
+    let ss = SyntaxSet::load_defaults_newlines();
+    let ts = ThemeSet::load_defaults();
+    let theme = match tm {
```
CMDJojo, line 112: We have two (to me) identical theme matches, and loading of `SyntaxSet` and `ThemeSet`, might be nicer to not do that in two places

packages/code/src/main.rs
```diff
@@ -94,13 +94,113 @@ fn transform_code(to: &str) {
...
+    };
+
+    if let Some(syntax) = ss.find_syntax_by_token(lang) {
+        let mut h = HighlightLines::new(syntax, theme);
+        let mut result = String::new();
+        let background_color = theme.settings.background.unwrap();
+        let r = background_color.r;
+        let g = background_color.g;
+        let b = background_color.b;
+        result.push('[');
```
CMDJojo, line 130: I think it may be error prone to construct arrays manually, maybe use `Vec<JsonEntry>` or `Vec<String>` or something and let `serde_json` serialize it?

packages/code/src/main.rs
```diff
@@ -94,13 +94,113 @@ fn transform_code(to: &str) {
...
+        write!(
+            result,
+            r#"{{"name": "raw", "data": "\\definecolor{{background}}{{RGB}}{{{r},{g},{b}}}\n"}},"#
+        )
+        .unwrap();
+        write!(
+            result,
+            r#"{{"name": "raw", "data": "\\begin{{center}}\n\\texttt{{\n\\begin{{codebox}}[colback=background]\n"}},"#
```
CMDJojo, line 138: I don't really think it is needed to use `center` here; it seems strange to do layout things here since it may (possibly, not tested) be incompatible with other layout packages such as `[row]`, `[center]` etc. I think you did it to make it start on a new line, but I think just a blank line `\\` would suffice. Or is there any other reason I don't think about?

packages/code/src/main.rs
```diff
@@ -94,13 +94,113 @@ fn transform_code(to: &str) {
...
+        .unwrap();
+        result.push(']');
+        result
+    } else {
+        eprintln!("Invalid language: {lang}");
+        code.to_string()
+    }
+}
+
+fn escape_latex_text(text: String, tab_size: u64) -> String {
```
CMDJojo, line 184: Do you need this? It seems to me like you can just use the `__text`, so that if we find something else that needs to be escaped, we can change it in one place, in the document template. Might be fine to keep anyways, since it seems like you have added some custom things here...

packages/code/src/main.rs
```diff
@@ -94,13 +94,113 @@ fn transform_code(to: &str) {
...
+        .replace('#', r"\#")
+        .replace('$', r"\$")
+        .replace('%', r"\%")
+        .replace('&', r"\&")
+        .replace('_', r"\_")
+        .replace('<', r"\textless{}")
+        .replace('>', r"\textgreater{}")
+        .replace('~', r"\textasciitilde{}")
+        .replace('^', r"\textasciicircum{}")
+        .replace("    ", &format!(r"\hspace*{{{tab_size}mm}}"));
```
CMDJojo, line 199: I think you misinterpreted how `tab_size` should be used. To my understanding, if you find a tab character (`\t`), you should replace it with `tab_size` amount of spaces (in width). I may have misundestood this, but it seems like you are replacing any occurance of four spaces `    ` with `tab_size` millimeters of horizontal space.

CMDJojo (changes requested):
Hi again!

Good job with your PR, it looks better now with the monospaced font and everything. I don't quite think everything is fully done yet, I have commented some in the code and moreover I have a few things to add:

* I have noted that spaces at the start of lines doesn't work, or multiple spaces in a row. I think it is very important to get fixed, since that is often used in code.
* It also seems to be a trailing newline at the end of the code block in LaTeX
* I would suggest formatting the code with `cargo fmt` so it looks a bit neater
* It would be nice to have a simple test or something just so that we know the module doesn't generate invalid JSON
* You have quite many commits, some of them not having the correct prefix. If you want, you may squash some of them to get a neater history

Using this code:
```
[code]
a
 b
  c
   d   e
```
gives this result:
image

adelhult:
I seem to encountered a strange issue, do you have any idea what may cause this?

When using latex all whitespace after `:` and a few other characters are stripped when using `lang=rust` but it seems to work fine for `lang=python` for instance.
image

image

It seems like the issue is only present when using latex:
image

CMDJojo:
Hmm. Might be worth checking out the built-in LaTeX-highlighter from Syntect? https://docs.rs/syntect/latest/syntect/util/fn.as_latex_escaped.html

Issue #215, opened 2023-04-11
Implement a scheduler which schedules transforms based on a graph generated by dependencies


CMDJojo:
We should not evaluate elements outermost-first or innermost-first like we did before, but rather have a scheduler which we can ask which element to evaluate next. This scheduler will ensure that a write to a variable occurs before a read to that variable.

PR #218, opened 2023-04-11
Add Scheduler to Core

CMDJojo:
This PR adds a new struct `Schedule` to `Core`. This struct keeps track of all elements there are and may be polled to get the next element it think should be evaluated. 

The code we had previously to evaluate a document was this:

```rs
pub fn eval_elem<T, U>(root: Element) -> Result<String, CoreError> {
    use Element::{Compound, Module, Parent};
    match root {
        Compound(children) => {
            let mut raw_content = String::new();

            for child in children {
                raw_content.push_str(&eval_elem(child, ctx, format)?);
            }
            Ok(raw_content)
        }
        Module { .. } | Parent { .. } => {
            let either = ctx.transform(&root, format)?;
            match either {
                Left(elem) => eval_elem(elem, ctx, format),
                Right(res) => Ok(res),
            }
        }
    }
}
```
*Note that some details not important to getting the point across are omitted from this and upcoming code snippets*

One can clearly see that it evaluates all elements from the outer element inwards. So, if we want to evaluate a `Parent` (like `__document`), we evaluate it and then recursively evaluate all new elements. This works fine if there are no dependencies, but if we have a dependency saying like "we need to evaluate all headings before the TOC" or "we need to compute all imports before adding the document header", this doesn't suffice.

The `Schedule`r knows about such dependencies, and instead of naively evaluate from the outside in, we poll the `Schedule`r for what element to evaluate next:
```rs
pub fn evaluate_scheduled<T, U>(mut root: Element) -> Result<String, CoreError> {
    let mut schedule = Schedule::default();
    schedule.add_element(&root, &ctx);
    while let Some(id) = schedule.pop() {
        let elem = root.get_by_id(id.clone()).unwrap();
        let new_elem = ctx.transform(&elem, format)?;
        schedule.add_element(&new_elem, &ctx);
        *root.get_by_id_mut(id).unwrap() = new_elem;
    }

    root.flatten().map(|s| s.join("")).unwrap()
}
```

For this to work, the `Either<Element, String>` structure was removed in favour of adding a new `Element` kind `Raw` which holds the evaluated string which will in the end be put in the document. This is because we need to be able to serialize it; previously we always evaluated parents first so no children were evaluated but this is just not always the case anymore. Additionally, we need a way to index elements, and I made the crate [granular-id](https://docs.rs/granular-id) for this purpose. It assigns the first element id `0`, the second id `1` etc, and all children of the first element will be assigned `0.0`, `0.1` etc.

The scheduler itself doesn't inherently know all dependencies, but when new elements are added, the context will be queried for what dependencies that specific element has. Currently, since we have no way to specify dependencies, there is an empty placeholder function which always returns that there is no dependencies.

To actually be able to write the code that generates the graph, some of the structure of variable declarations and dependency types, orderings etc were needed. For this reason, `variables.rs` was created, holding such enums. Note that this **does not implement variables**, it just has some enums for different variable types, and these (as well as the access types) may be modified when actually implementing the variable store.

This PR is to a large extent done. There is some clean-up work left to do before it is ready for merging. If you want to test this out, you can hard-code the `get_variable_accesses` function like this:
```rs
    pub fn get_variable_accesses(&self, name: &str) -> Option<Vec<(Variable, VarAccess)>> {
        match name {
            "__paragraph" => Some(vec![(Variable("abc".to_string(), Constant), VarAccess::Constant(ConstantAccess::Read))]),
            "__heading" => Some(vec![(Variable("abc".to_string(), Constant), VarAccess::Constant(ConstantAccess::Declare))]),
            _ => Some(vec![])
        }
    }
```
This more or less says that any headings must be evaluated before any paragraphs. Then, you can add prints to the `transform_from_wasm` function to see what elements are being transformed, and test with some documents containing multiple interleaved paragraphs and headings and see that the headings are in fact evaluated before the paragraphs.

Resolves GH-215.

### DIFF
--- a/core/src/element.rs
+++ b/core/src/element.rs
@@ -24,11 +24,12 @@ pub enum Element {
         id: GranId,
     },
     Compound(Vec<Self>),
+    Raw(String),
 }
 
 impl Element {
     pub fn get_by_id(&self, id: GranId) -> Option<Self> {
-        let components: Vec<u32> = id.into();
+        let components: Vec<usize> = id.into();
         components
             .into_iter()
             .fold(Some(self), |current, id| {
@@ -42,7 +43,7 @@ impl Element {
     }
 
     pub fn get_by_id_mut(&mut self, id: GranId) -> Option<&mut Self> {
-        let components: Vec<u32> = id.into();
+        let components: Vec<usize> = id.into();
         components.into_iter().fold(Some(self), |current, id| {
             current.and_then(|c| match c {
                 Element::Parent { children, .. } => children.get_mut(id as usize),
@@ -52,30 +53,26 @@ impl Element {
         })
     }
 
-    /*fn with_id(mut self, new_id: GranId) -> Self {
+    pub fn is_flat(&self) -> bool {
         match self {
-            Element::Parent(ref mut id, _, _) => *id = new_id,
-            Element::Module(ref mut id, _) => *id = new_id,
-            Element::Compound(ref mut id, _) => *id = new_id,
-            Element::Raw(ref mut id, _) => *id = new_id,
+            Element::Parent { .. } | Element::Module { .. } => false,
+            Element::Compound(c) => c.iter().all(Element::is_flat),
+            Element::Raw(_) => true,
         }
-        self
-    }*/
+    }
 
-    /// Attempt to flatten the element by merging raw elements and compounds
+    /// Attempt to flatten the element by merging raw elements and compounds. If any other element
+    /// is in the structure, this function returns `None`
     pub fn flatten(self) -> Option<Vec<String>> {
         match self {
-            Element::Compound(children) => children.into_iter().map(Self::flatten).fold(
-                Some(Vec::new()),
-                |mut vec, mut flat| {
-                    if let Some(ref mut v) = flat {
-                        vec.as_mut().map(|x| x.append(v));
-                    }
-                    vec
-                },
-            ),
-            // TODO: Add a raw kind
-            Element::Raw(_, value) => Some(vec![value]),
+            // Note that the collect ensures we have an early return if any of the elements can't
+            // be flattened
+            Element::Compound(children) => children
+                .into_iter()
+                .map(Self::flatten)
+                .collect::<Option<Vec<Vec<String>>>>()
+                .map(|x| x.into_iter().flatten().collect()),
+            Element::Raw(s) => Some(vec![s]),
             // Parent and module nodes can't be flattened and must be evaluated
             _ => None,
         }
@@ -91,7 +88,7 @@ impl Element {
         }
     }
 
-    pub(crate) fn try_from_ast(value: Ast, id: GranId) -> Result<Self, CoreError> {
+    pub fn try_from_ast(value: Ast, id: GranId) -> Result<Self, CoreError> {
         macro_rules! zip_elems {
             ($elems:expr, $id:expr) => {
                 $elems
--- a/core/src/element.rs
+++ b/core/src/element.rs
@@ -34,8 +34,8 @@ impl Element {
             .into_iter()
             .fold(Some(self), |current, id| {
                 current.and_then(|c| match c {
-                    Element::Parent { children, .. } => children.get(id as usize),
-                    Element::Compound(children) => children.get(id as usize),
+                    Element::Parent { children, .. } => children.get(id),
+                    Element::Compound(children) => children.get(id),
                     _ => None,
                 })
             })
@@ -46,8 +46,8 @@ impl Element {
         let components: Vec<usize> = id.into();
         components.into_iter().fold(Some(self), |current, id| {
             current.and_then(|c| match c {
-                Element::Parent { children, .. } => children.get_mut(id as usize),
-                Element::Compound(children) => children.get_mut(id as usize),
+                Element::Parent { children, .. } => children.get_mut(id),
+                Element::Compound(children) => children.get_mut(id),
                 _ => None,
             })
         })
--- a/core/src/element.rs
+++ b/core/src/element.rs
@@ -1,5 +1,7 @@
 use std::collections::HashMap;
 
+use granular_id::GranularId;
+
 use parser::{Ast, MaybeArgs, ModuleArguments};
 
 use crate::CoreError;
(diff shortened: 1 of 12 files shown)


core/src/schedule.rs
```diff
@@ -0,0 +1,180 @@
...
+
+// In short, each element is identified by its "GranId". We can use these Ids in our sorting, and
+// then get the next ID to be evaluated. But when a parent node is evaluated, all its children
+// is invalid and any returned IDs from its children (which is "new") may previously have been
+// assigned to the previous children (which doesn't exist anymore). This means that we need to be
+// able to "remove" IDs, but that isn't possible in TopologicalSort. For this reason, we assign
+// new unique IDs to each GranId, of the type ScheduleId. We have an BiBTreeMap to keep track of
+// this mapping.
+
+pub struct Schedule {
```
adelhult, line 23: Not sure about this once, but we may not want to expose the schedule as part of the public API and make these `pub(crate)` instead. Or maybe the schedule module is never exposed anyway.

core/src/schedule.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub fn pop(&mut self) -> Option<GranId> {
+        // TODO LATER: use pop all and reorder them so modules of the same type are evaluated
+        //  after each other, that way we can keep the same wasm module alive for better performance
+
+        // schedule_id is the ID of the next node to be evaluated
+        while let Some(schedule_id) = self.dag.pop() {
+            // If this isn't in our BiBTreeMap it means that it doesn't exist anymore (it may once
+            // have been a child of an element that is now evaluated). Then we pop the next ID.
+            let Some((gran_id, _)) = self.id_map.remove_by_right(&schedule_id) else {
+                continue;
```
adelhult, line 55: Mostly a nit-pick but it can be kind of nice to move the comment to the relevant clause.
```suggestion
            let Some((gran_id, _)) = self.id_map.remove_by_right(&schedule_id) else {
                 // If this isn't in our BiBTreeMap it means that it doesn't exist anymore (it may once
                 // have been a child of an element that is now evaluated). Then we pop the next ID.
                continue;
```

core/src/schedule.rs
```diff
@@ -0,0 +1,180 @@
...
+            };
+
+            // We need to remove all children, and we do this by getting the next sibling and
```
adelhult, line 58: Nice comments!

core/src/schedule.rs
```diff
@@ -0,0 +1,180 @@
...
+                // This may be optimised with other data structures
+                self.dep_info.iter_mut().for_each(|(_k, v)| {
+                    v.retain(|elem| elem.0 != id);
+                });
+            });
+            return Some(gran_id);
+        }
+
+        // In this case, we couldn't pop one element, and this may occur either if there is a
+        // cycle or if the DAG is empty. This will be checked externally.ﬁ
```
adelhult, line 86:
```suggestion
        // cycle or if the DAG is empty. This will be checked externally.
```

core/src/schedule.rs
```diff
@@ -0,0 +1,180 @@
...
+        if let Element::Compound(children) = element {
+            for child in children {
+                self.add_element(child, ctx, format);
+            }
+            return;
+        }
+
+        // For adding to the schedule, we need the name and GranId for that element
+        // We could do this with a guard (let {name, id} = element else ...) but then rustfmt
+        // doesn't format it for some reason
```
adelhult, line 112: We can remove this comment, since it's absolutely fine to use a match

core/src/schedule.rs
```diff
@@ -0,0 +1,180 @@
...
+        &mut self,
+        element: &Element,
+        ctx: &Context<T, U>,
+        format: &OutputFormat,
+    ) {
+        // Element may be one of four kinds:
+        // * Raw => We don't add it to the schedule
+        // * Module => We add it to the schedule
+        // * Parent => We add it and all its children to the schedule
+        // * Compound => We add all its children to the schedule, but not the compound itself
```
adelhult, line 100: I think that it might be nice to refactor this function in the way that the comment describes with a match expression instead of all the early returns.

core/src/variables.rs
```diff
@@ -0,0 +1,96 @@
...
+    }
+}
+
+impl PartialOrd for VarAccess {
+    // Two different accesses `a` and `b` may or may not need to be ordered in a certain way.
+    // The result when comparing the two determines how they are ordered:
+    // * If `a` < `b`, `a` *will* occur before `b`
+    // * If `a` > `b`, `a` *will* occur after `b`
+    // * If `a` = `b`, `a` may occur before or after `b` (see ::order_granular)
+    // * If there is no ordering (this function returns None), any may occur before the other
```
adelhult, line 41:
```suggestion
    /// Two different accesses `a` and `b` may or may not need to be ordered in a certain way.
    /// The result when comparing the two determines how they are ordered:
    /// * If `a` < `b`, `a` *will* occur before `b`
    /// * If `a` > `b`, `a` *will* occur after `b`
    /// * If `a` = `b`, `a` may occur before or after `b` (see ::order_granular)
    /// * If there is no ordering (this function returns None), any may occur before the other
```

core/src/variables.rs
```diff
@@ -0,0 +1,96 @@
...
+pub enum VarAccess {
+    Set(SetAccess),
+    List(ListAccess),
+    Constant(ConstantAccess),
+}
+
+impl VarAccess {
+    // If two access levels are deemed the same (no one should strictly be done before the other),
+    // this function is called to see if the granularid (occurance in the source document) should
+    // determine the order of evaluation. If not, the order is arbitrarily chosen.
```
adelhult, line 29:
```suggestion
    /// If two access levels are deemed the same (no one should strictly be done before the other),
    /// this function is called to see if the granularid (occurance in the source document) should
    /// determine the order of evaluation. If not, the order is arbitrarily chosen.
```

core/src/variables.rs
```diff
@@ -0,0 +1,96 @@
...
+        }
+    }
+}
+
+// Note we can have two different variables with the same name
+// if they have different types
+#[derive(Clone, Debug, Hash, Eq, PartialEq)]
+pub struct Variable(pub String, pub VarType);
+
+// The ordering of these is very important. For determining what variable access types must occur
```
adelhult, line 75:
```suggestion
// The ordering of these enum variants is very important. For determining what variable access types must occur
```

adelhult (commented):
Looks great! Left a bunch of comments, mostly regarding docs though so I think it won't take long before we can rebase.

core/src/element.rs
```diff
@@ -1,29 +1,106 @@
...
+impl Element {
+    pub(crate) fn id(&self) -> Option<&GranId> {
+        if let Element::Parent { id, .. } | Element::Module { id, .. } = self {
+            Some(id)
+        } else {
+            None
+        }
+    }
+
+    pub fn try_from_ast(value: Ast, id: GranId) -> Result<Self, CoreError> {
```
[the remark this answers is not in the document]
CMDJojo: This is only used when parsing, and won't need to be a part of the public api. My bad not making it pub(crate)! I did that now (and added a comment as well)

core/src/schedule.rs
CMDJojo: Won't be good, discussed in private with you

core/src/schedule.rs
CMDJojo: I think the schedule module is never exposed, but I will make it pub(crate)!

playground/web_bindings/Cargo.toml
```diff
@@ -25,6 +25,7 @@ wasmer-vfs = { version = "3.1.1", default-features = false, features = ["mem-fs"
 js-sys = "0.3.61"
 wasm-bindgen-futures = "0.4.34"
 once_cell = "1.17.1"
+granular-id = "0.4.1"
```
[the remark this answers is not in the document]
CMDJojo: Yes, absolutely!

CMDJojo:
Thank you for your review @adelhult ! I have now changed the things you commented on. Nice idea promoting some of the comments to doc comments, didn't actually think of that myself!

adelhult (approved):

Issue #192, opened 2023-03-27
Add references


hugomardbrink:
Support references in the document in the link package

PR #219, opened 2023-04-11
GH-192: Add internal referencing

hugomardbrink:
- Resolves: GH-192.

### DIFF
--- a/packages/link/src/main.rs
+++ b/packages/link/src/main.rs
@@ -2,7 +2,6 @@ use std::env;
 use std::io::{self, Read};
 
 use serde_json::{json, Value};
-
 fn main() {
     let args: Vec<String> = env::args().skip(1).collect();
     let action = &args[0];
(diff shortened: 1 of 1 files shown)


packages/link/src/main.rs
```diff
@@ -100,3 +113,100 @@ fn transform_link(to: &str) {
...
+            let label_tag = format!(r#"\label{{{}}}"#, escaped_label);
+
+            let output = json!([
+                {"name": "raw", "data": label_tag},
+            ]);
+
+            print!("{output}");
+        }
+        other => {
+            eprintln!("Cannot convert ref to {other}");
```
adelhult, line 157:
```suggestion
            eprintln!("Cannot convert label to {other}");
```

packages/link/src/main.rs
```diff
@@ -100,3 +113,100 @@ fn transform_link(to: &str) {
...
 }
+
+fn transform_label(to: &str) {
+    match to {
+        "html" => {
+            let input: Value = {
+                let mut buffer = String::new();
+                io::stdin().read_to_string(&mut buffer).unwrap();
+                serde_json::from_str(&buffer).unwrap()
+            };
```
adelhult, line 124: Maybe move this outside of the match statement  to avoid repeating it twice

adelhult (approved):
Great work! Left comments with a few nit-picks

Issue #197, opened 2023-03-29
Fix y-overflow in playground


Forsinge:
**Describe the bug**
Two scrollbars appear around the output when opening playground on Windows.

*Update: I forgot to include which browser I was running. In the screenshots I am running Chrome (probably version 111).*  

**Screenshots**
![image](https://user-images.githubusercontent.com/89988910/228466593-7bbb45ae-ac06-4b31-813f-fc52d53ec51c.png)

CMDJojo:
There is also a scrolling issue, possibly related to this, which happens on my iPhone when opening the packages view. When I scroll to the bottom of the package list, I don't get to the bottom and the last package is below the screen. If I wait and continue scrolling, I can "scroll away" the taskbar and thus see the last package. I am not that good at CSS to determine if this is the same cause or not. Attaching screen recording below.

PR #221, opened 2023-04-13
GH-197: Fix y-overflow in playground

Forsinge:
Resolves: #197 

It seems the issue was caused by overwriting the overflow property to "auto" for the editor element. Ace sets this to "hidden" otherwise.

### DIFF
--- a/playground/static/style.css
+++ b/playground/static/style.css
@@ -145,13 +145,13 @@ select:hover,
     width: 50%;
     height: 100%;
     box-sizing: border-box;
-    overflow: auto;
     margin: 0;
     border: none;
     transition: 0.3s width;
 }
 
 #preview {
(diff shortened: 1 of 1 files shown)


CMDJojo (approved):
Looks good to me! Note that this doesn't fix the issue on iOS also linked in the original issue, but that probably has another root problem (at least according to @adelhult when I spoke to him, I am not very good at CSS)

adelhult:
> Looks good to me! Note that this doesn't fix the issue on iOS also linked in the original issue, but that probably has another root problem (at least according to @adelhult when I spoke to him, I am not very good at CSS)

Yeah, I think we can leave that other issue for another day. I suspect that we will go back and improve the design of the playground a bit more before the final presentation anyhow : )

Issue #211, opened 2023-04-06
Infinite recompilation loop when using live preview


adelhult:
**Describe the bug**
When attempting to use the live preview in the bachlor-thesis repo I noticed that the CLI gets stuck in a infinite loop of recompilations. This seems to be  due to the debounce file watcher not caring about if we read or write to a file and it takes us much more then the debounce period to read all the extra tex files in the thesis. 

**Expected behavior**
The document should only reload once after a change in one of the mdm files

**Screenshots**
![image](https://user-images.githubusercontent.com/11508459/230446333-a763d203-708b-43ba-b79a-507e1d5dd1ff.png)

PR #226, opened 2023-04-14
Remove debounce watcher

adelhult:
Resolves GH-211 by removing the debouncing file watcher in favour of a simpler one. A bit of regression since I think we are at risk of doing multiple unnecessary recompilations back-to-back in some cases. But, it should not be to bad and I rather have something semi-optimal that works then something broken.

### DIFF
--- a/cli/Cargo.toml
+++ b/cli/Cargo.toml
@@ -21,8 +21,8 @@ tokio = { version = "1.26.0", features = ["full"] }
 tokio-stream = "0.1.12"
 warp = "0.3.3"
 futures-util = "0.3.27"
-notify-debouncer-mini = { version = "0.2.1", default-features = false }
 portpicker = "0.1.1"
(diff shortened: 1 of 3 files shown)


axoxelol1 (commented):
Watcher works both with the web live preview and outputting to file for the thesis now, nice work 👍 I sometimes have to refresh the website after change as it simply goes blank but I am not able to recreate it consistently, is this anything you have experienced?

adelhult:
Yeah, sadly I've encountered that before too, that's why I implemented the debounce watcher. I had trouble recreating the bug when I made this PR so I assumed that I had gotten away with it. 

I pushed a new update that sleeps for 20ms between each recompilation hopefully that will fix it? Perhaps you could try it out.

(I think the bug is caused due to the filer still being written to when the watcher is triggered and starts the compilation).

axoxelol1 (approved):
I am not able to get the bug to occur now so hopefully the short sleep fixed it. Like I said in a previous comment everything seems to work as it should =)

Issue #227, opened 2023-04-14
Add argument in image module to support alt text


Forsinge:
**Is your feature request related to a problem? Please describe.**
Alt text is an important addition to improve accessibility. A possible implementation is adding it as the first argument, and letting it be optional but giving a warning when empty.

PR #228, opened 2023-04-14
GH-227: Add alt-text to image

CMDJojo:
This gives a compilation warning if an alt-text isn't given. It also escapes quotes and other characters that should be escaped in HTML (using `__text`)

Resolves GH-227


Forsinge (approved):
Looks good!

Issue #203, opened 2023-04-02
Better naming of package path `specifier:`s


adelhult:
When importing a package the user does so with a path that starts with a specifier (terminology which is borrowed from the Deno JS runtime). Here are a few examples:
* A local package relative to the assets dir `foo/bar.wasm` or `foo/bar`
* A package on the web `https:/some.website/package.wasm`
* A package found in the registry (which we currently call "pkgs", analogous to npm or ctan) `pkgs:bar`
* A package that is baked into the compiler `std:latex

**It is worth discussing if the name of the registry `pkgs` should be changed into something else to make it clear that this is just the name of the registry and not just the way to import packages. Possible alternatives would be `reg:` or `registry` or rename the registry into something less confusing.**

We might also want to discuss if "std" should be called something else or if we want to introduce a `prelude:` specifier.

adelhult:
We opted to use `catalog:` instead!

PR #230, opened 2023-04-14
Rename "registry" to "catalog"

adelhult:
Resolves GH-203 by renaming registry into catalog and switch the `pkgs:` specifier to `catalog:`


Forsinge (approved):
Looks good!

Issue #217, opened 2023-04-11
Add a variable store which keeps track of all variables and checks for errors in runtime


CMDJojo:
We should have a variable store that keeps track of all variables there are, allows reads and writes to/from them, and possibly add some error checking into it (like a flag for constants, so that if they are ever read, a warning is emitted if a write occurs afterwards)

PR #231, opened 2023-04-14
Implement variable store

adelhult:
This PR implements:
* A `VariableStore` which lives in the `Context` struct and keeps all shared state.
* Six new modules: `const-read`, `const-decl`, `list-push`, `list-read`, `set-add`, `set-read` for manipulating variables stored in the shared state.
* A way of accessing the variables in packages by using WASI env variables `env::var("<TYPE>_<NAME>=<VALUE>")`.

Feel free to try out, I've left a sample document at the bottom of the PR.

# Questions
* Currently we allow two variables to have the same name as long as they have different types. This is a bit unconventional and complicates the code somewhat. And most importantly it might cause some confusions for package authors because if they declare a variable `"variables: { "foo": {type: "const", access: "read"}}` they must remember to use key `const_foo` when accessing it and not just `foo`. **Should we keep this behaviour or not?**
* Sadly `-` is now allowed in env variables. This means that it would be impossible to access a variable named `some-thing` for example. Either we keep it this way and give a error if the user attempts to create a variable containing "-". Alternatively, we could treat "-" and "_" as the same character and convert every hyphen to underscore when sending the env vars using WASI, but this might end up confusing users, **what do you think we should do?**
* Naming of the modules, feel free to suggest any alternative names too. For instance, I'm not convinced that `const-decl`  is the best name...

## Example document
```
# Contents

[list-read headings]()

# Figures
[list-push headings](Figures)
[set-read figures]{}

# All figures
[list-push headings](All figures)

# Intro
[list-push headings](Intro)

# Abstract
[list-push headings](Abstract)


[code matlab theme=github]
hold on
plot(x, y);


[code py theme=eighties]
for x in range(10):
    print(x)

[code py]
for x in range(10):
    print(x)


[const-decl code_theme](eighties)

[set-add figures](1)
[set-add figures](1)
[set-add figures](1)
[set-add figures](2)
[set-add figures](3)
[set-add figures](4)
[set-add figures](5)
[set-add figures](6)
[set-add figures](7)
```

Resolves GH-217

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -16,14 +16,14 @@ use wasmer::{Cranelift, Engine, EngineBuilder};
 use wasmer::{Instance, Module, Store};
 use wasmer_wasi::{Pipe, WasiState};
 
-use parser::config::{Config, HideConfig, ImportConfig};
+use parser::config::{self, Config, HideConfig, ImportConfig};
 use parser::ModuleArguments;
 
 use crate::element::GranularId;
 use crate::fs::CoreFs;
 use crate::package::{ArgValue, PackageImplementation};
 use crate::package_store::{PackageID, PackageStore};
-use crate::variables::{VarAccess, Variable};
+use crate::variables::{VarAccess, VarType, VariableStore};
 use crate::CoreError::MissingTransform;
 use crate::{std_packages, AccessPolicy, Element, Resolve};
 use crate::{ArgInfo, CoreError, OutputFormat, Package, Transform};
@@ -36,12 +36,13 @@ pub struct Context<T, U> {
     pub(crate) state: CompilationState,
     pub filesystem: CoreFs<U>,
     policy: Arc<Mutex<U>>,
-    // This is a temporary field for testing variables
-    pub(crate) lists: HashMap<String, Vec<String>>,
 }
 
+/// Contains volatile compilation state that should be cleared
+/// in between calls to evaluation functions
 #[derive(Default, Clone, Debug)]
 pub struct CompilationState {
+    pub variables: VariableStore,
     pub warnings: Vec<Issue>,
     pub errors: Vec<Issue>,
     pub verbose_errors: bool,
@@ -77,6 +78,7 @@ impl CompilationState {
     fn clear(&mut self) {
         self.warnings.clear();
         self.errors.clear();
+        self.variables.clear();
     }
 }
 
@@ -144,7 +146,6 @@ impl<T, U> Context<T, U> {
             state: CompilationState::default(),
             filesystem: CoreFs::new(Arc::clone(&policy)),
             policy,
-            lists: HashMap::new(),
         };
         #[cfg(feature = "native")]
         ctx.package_store
@@ -175,6 +176,19 @@ where
         #[cfg(feature = "web")]
         store_guard.register_resolved_packages()?;
 
+        // Declare any constants that were specified in the [config] module
+        let declaration_errors: Vec<CoreError> = config
+            .sets
+            .iter()
+            .filter_map(|config::Set { key, value }| {
+                self.state.variables.constant_declare(key, value).err()
+            })
+            .collect();
+
+        if !declaration_errors.is_empty() {
+            return Err(declaration_errors);
+        }
+
         let arc_mutex = Arc::clone(&self.package_store);
         let resolve_tasks = store_guard.generate_resolve_tasks(arc_mutex, &config)?;
         if resolve_tasks.is_empty() {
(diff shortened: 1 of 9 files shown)


axoxelol1:
My thoughts on the questions:
- I feel like allowing multiple variables with the same name can be confusing, I don't really see a point in having that behavior.
- Converting `-` to `_` could also be very confusing as users might expect `some-var` and `some_var` to be different variables but they both point to the same one. It is unfortunate that `-` isn't allowed but it might be easiest to just use a snake case naming convention for variables then.
- Naming is always difficult, `decl const` and `read const` might be easier to understand but it is nice to have the same convention for all modules eg `verb-type` and changing `set-add` to `add-set` makes it sound like a type of set and is confusing.

Forsinge:
I think I agree with what @axoxelol1 said. I don't have much to add to his answer. In general I'm not too keen on hiding behaviour from the user. Also, if we don't allow any duplicate names, having a single `[read]` module seems quite nice to me.

CMDJojo:
Good job with the PRs!

I think Axels comments are good, but personally I prefer `type-verb` like `const-declare` and `list-append` since I think the type is more "important" and should be more prominent in the name than the actual operation. It also makes more sense when thinking of things ("I have a list, and I want to append to it" rather than "I want to append something, and that something is a list"). For auto-completion, the type also makes more sense imo since you probably think of the variable you want to touch before how, similar like if you have a variable you know the type before what the actual operation is (you don't think "I want to run the prefix function on something, and that something is a vec", rather "I have a vec, I want the prefix function on that").

Additionally, I would argue for a new variable type `ChainedCounter` (or change our current counter to have that behavior). What I think of is a counter which have multiple numbers where you may advance any one of them. You may choose to advance any level, clearing all lower levels. This would allow us to add headers like `1`, `1.3.2` and so for headings (where if you make a top-level you advance the level 1 component and read it, if you make a level 2 heading, you advance the second component and read it etc). Our current types doesn't allow for this if we don't make one counter for each level and reset them all manually, which would make us need to overload names (have variables `counter_heading1`, `counter_heading2` etc), and have to keep track of the number of levels. (This would be similar to GraunlarId but maybe starting at 1).

Then, it is important to think of accesses. It seems like the main way to access and use a variable from within a module is by `env`, but what happens if a constant is undeclared? I guess it doesn't appear in env and that is reasonable, but what about sets? It seems more reasonable to return an empty set if nothing has been pushed to it rather than leaving it undeclared, since that would leave one less check for the developer. How is the set values encoded? Maybe JSON encoding would be best here since that has escaped characters, works well with libraries, and would allow all strings to be encoded. If we do it manually and use like a comma as separator, we need to escape that, and escape the escape character etc etc. JSON seems like a good solution for this. The same things can be applied to lists; JSON encoding and empty list as the "undefined" value.

One other type that may be debatable to add is `Variable` that is a variable that can be changed and has the value it was most recently set to. This would allow for for-loops:
```
[repeat 3 counter]
[var-read counter]()
```
could expand to
```
[var-set counter] 1
[var-read counter]()
[var-set counter] 2
[var-read counter]()
[var-set counter] 3
[var-read counter]()
```
I do not know if this is a good idea.

These are my current thoughts. Feedback is appreciated!

adelhult:
Oops, sorry. I meant hyphens @CMDJojo. Btw, @Forsinge I liked your suggestion of using a single read module but it seems quite difficult to implement with our current setup

CMDJojo:
Nice job with the PR!

Since you asked for a review, I can give you some feedback. First of all, I made some suggestions in a previous comment, but just looking briefly at your code, I found some additional things that may be worth changing.

First of all, I think the general way you do lookups and such in the map (and the map itself) is kind of off. You have everything in a single `HashMap<(String, VarType), Value>`, which I think can be done better.
* First of all, you want to be able to index this with `(&String, &VarType)`, but that isn't possible, so you do a weird workaround involving trait objects, dynamic dispatch and such. I think it looks very confusing, complicated and reduces performance due to dynamically dispatched function calls and such.
* Secondly, we are encoding the type in two different ways, which may have collisions. You have the type encoded separately in both the key and the value, so you do a lookup on `(name, type)`, and expect an enum variant matching the type as the result. In general, it is better to leverage the type system to make it safer and ensure that there can't be any invariants that may be broken.

If we change your type from just holding a single `HashMap<(String, VarType), Value>` to containing three maps like this:
```
pub struct VariableStore {
    constants: HashMap<String, String>,
    lists: HashMap<String, Vec<String>>,
    sets: HashMap<String, HashSet<String>>
}
```

If we do this, we get these benefits:
* We don't need any dynamic dispatch for accesses
* We can have a function `get_variable` which takes an `(&str, &VarType)` (akin to indexing by that, which is what you wanted in the first case, that you did the dynamically dispatched workaround to achieve) that gets the variable
* We don't have that trait which is mostly a workaround, may be hard to understand and seems unnecessary
* We leverage the type system; if we query a `(&str, VarType::Set)` we know that we will get a set and don't have to double-check it
* When doing lookups in native modules, this means that we can skip the step where we wrap it into a `Value` just to unwrap it again and panic if the type is mismatched; it is just not possible for a type mismatch. This also improves performance of that code
* We get better performance by allowing the Rust compiler to possibly optimize the polymorphism due to using a more specific type
* We may get a smaller binary size and faster compilation if HM is already polymorphed to that type
* The HM itself gets smaller due to not needing to store the type information in the keys nor values

In addition, it seems like you changed the inner workings for `get_var_dependencies` (previously `get_variable_accesses`) so that it no longer fails on Compound or Raw elements - I think it is good if it does fail on Compound ad Raw elements since we know that this eventually leads to problems during compilation. It seems like you completely rewritten the function, I don't really understand why since I think my implementation was sufficiently good and I don't see anything that gets better with the rewrite, just some functionality that got removed, but yeah. It is no longer fail-fast, possibly making logic errors in the scheduler very hard to spot, and in some cases it may lead to problems not visible expect in some strange circumstances, based on RNG, that will be very **very** hard to debug. Is there any good reason behind this change?

Anyways, as said, good job. Will do a proper review later on, when some of my previous comments have been answered/resolved.

adelhult:
Yeah, It was probably good of you to suggest to remove the `AsVariable` trait. I also tried to implement the VariableStore using multiple maps however when I refactored the code to take @axoxelol1 input into account and ensure that we won't get two variables with the same name I think it ended up cleaner using a single map.

New changes based on your input:
* I've reverted the change in `get_var_dependencies`
* Constants can now be declared in the `[config]` module using the `set NAME VALUE` syntax. (I can open a issue later for changing the naming of either "const-decl" or "set" later. But it's  was quite nice to be able to declare constants at the top-level : )
* I added a comment about the fact that lists and sets are encoded in JSON. Nice that you noticed that was missing.
* I re-factored quite a lot of the code and introduced a `TypeMismatch` error so that we now act more alike to other language and don't allow multiple variable with different name:s

core/src/variables.rs
```diff
@@ -1,28 +1,175 @@
...
+    pub fn clear(&mut self) {
+        self.0.clear()
+    }
+
+    /// Ensure that a valid name is used.
+    /// Only ASCII alphanumerics characters and "_" is allowed. The name may also not start with a number.
+    fn valid_name(&self, name: &str) -> Result<(), CoreError> {
+        // Ensure that the first character is not a number and that the name
+        // is at least 1 character long
+        if name.chars().next().map(char::is_numeric).unwrap_or(true) {
```
CMDJojo, line 27: Here, you are creating the same iterator twice. It seems unneccecary. I think we can do somthing like this:
```rust
let mut chars = name.chars();
let valid = chars.next().map(|c| c.is_ascii_alphabetic() || c == '_').unwrap_or_default() && chars.all(|c| c.is_ascii_alpanumeric() || c == '_');
```
or something like that, to make sure we don't creat the iterator twice. Maybe doesn't have that large of an impact anyhow

core/src/variables.rs
```diff
@@ -1,28 +1,175 @@
...
+    }
+
+    /// Clears the variable store
+    pub fn clear(&mut self) {
+        self.0.clear()
+    }
+
+    /// Ensure that a valid name is used.
+    /// Only ASCII alphanumerics characters and "_" is allowed. The name may also not start with a number.
+    fn valid_name(&self, name: &str) -> Result<(), CoreError> {
```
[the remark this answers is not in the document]
CMDJojo: Also, maybe `valid_name` could be changed for a better name, like `validate_name`? `valid_name` sounds like `is_valid_name` which would return a bool.

core/src/std_packages.rs
```diff
@@ -174,41 +260,125 @@ pub fn native_block_content<T, U>(
...
+    let name = args.get("name").unwrap().as_str().unwrap();
+
+    match ctx.state.variables.get(name) {
+        Some(value @ variables::Value::Set(_)) => Ok(text_element(value.to_string(), id.clone())),
+        Some(value) => Err(CoreError::TypeMismatch {
+            name: name.to_string(),
+            expected_type: VarType::Set,
+            present_type: value.get_type(),
+        }),
+        None => Ok(Element::Compound(vec![])), // TODO: give a error or warning
```
CMDJojo, line 381: Something to do?

core/src/std_packages.rs
```diff
@@ -174,41 +260,125 @@ pub fn native_block_content<T, U>(
...
+    match ctx.state.variables.get(name) {
+        Some(value @ variables::Value::Constant(_)) => {
+            Ok(text_element(value.to_string(), id.clone()))
+        }
+        Some(value) => Err(CoreError::TypeMismatch {
+            name: name.to_string(),
+            expected_type: VarType::Constant,
+            present_type: value.get_type(),
+        }),
+        None => Ok(Element::Compound(vec![])), // TODO: give a error or warning
```
CMDJojo, line 309: Something to do?

core/src/std_packages.rs
CMDJojo: Out of our previous discussions, I have really liked the idea of defaulting to an empty list/set when reading something that hasn't been declared, while you have been very keen on keeping it undefined. I think the most common thing people would expect when doing `[set-read] imports` is not to give a big error but rather not print any authors if there aren't any. I know you disagree, but it might be a good idea to add an argument to default to an empty set if absent?

core/src/error.rs
```diff
@@ -102,6 +102,22 @@ pub enum CoreError {
...
+    },
+    #[error("Attempted to access the variable '{name}' as a '{expected_type}' but the name is already occupied by value of type '{present_type}'.")]
+    TypeMismatch {
+        name: String,
+        expected_type: VarType,
+        present_type: VarType,
+    },
+    #[error("Attempted to redeclare the constant '{0}'.")]
+    ConstantRedeclaration(String),
+    #[error("Forbidden variable name '{0}'. Only ASCII letters and numbers as well as '_' is allowed. The name may also not start with a number.")]
```
CMDJojo, line 119:
```suggestion
    #[error("Forbidden variable name '{0}'. Only ASCII letters and digits as well as '_' is allowed. The name may also not start with a number.")]
```

packages/code/src/main.rs
```diff
@@ -57,16 +60,43 @@ fn transform(from: &str, to: &str) {
...
+            if [
+                "ocean-dark",
+                "ocean-light",
+                "mocha",
+                "eighties",
+                "github",
+                "solar-dark",
+                "solar-light",
+            ]
+            .contains(&value.as_str())
```
CMDJojo, line 89: Maybe print a warning or something if the value isn't one of these?

packages/code/src/main.rs
```diff
@@ -36,11 +36,14 @@ fn manifest() {
...
                     {"name": "font_size", "default": 12, "description": "The size of the font", "type": "uint"},
                     {"name": "tab_size", "default": 4, "description": "The size tabs will be adjusted to", "type": "uint"},
-                    {"name": "theme", "default": "mocha", "description":
+                    {"name": "theme", "default": "default", "type": ["default", "ocean-dark", "ocean-light", "mocha", "eighties", "github", "solar-dark", "solar-light"], "description":
                         "Theme of the code section. For available themes, see \
                         https://docs.rs/syntect/latest/syntect/highlighting/struct.ThemeSet.html#method.load_defaults"},
                     {"name": "bg", "default": "default", "description": "Background of the code section"},
                 ],
+                "variables": {
+                    "code_theme": {"type": "const", "access": "read"}
```
CMDJojo, line 45: Maybe document somewhere (like on the theme argument, or transform descr) that this is used?

core/src/context.rs
```diff
@@ -192,112 +210,118 @@ where
...
+    fn get_vars_to_read(
+        &self,
+        element: &Element,
+        format: &OutputFormat,
+    ) -> Result<Vec<(String, VarType)>, CoreError> {
+        let variables = self
+            .get_var_dependencies(element, format)?
+            .into_iter()
+            .filter_map(|(variable, access)| {
+                if access.is_read() {
```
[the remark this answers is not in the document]
CMDJojo:
```suggestion
                access.is_read().then_some(variable)
```

CMDJojo (changes requested):
Good job with the PR! It is really nice to get variables up and working!

I found some rough edges in your code that I commented on. The reason I didn't do a review earlier is because many of the things I commented on that I still believe are issues that should be resolved before merging, whereas very few of the changes have been implemented, so I thought it would be best to wait until those changes were implemented. But anyways, I looked though the code a bit and commented on some things I thought could be changed, not very large things.

As said, I still believe we should address some of my previous comments before merging.

core/src/element.rs
```diff
@@ -77,6 +77,15 @@ impl Element {
             _ => None,
         }
     }
+
+    /// Get the name of a element (if it has one).
+    pub fn name(&self) -> Option<&String> {
```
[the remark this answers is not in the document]
adelhult: Changing it to `&str`. And yes it's a good point that it might be a bit unnecessary since I only use it once but it seems like a reasonable method nd it looks like it can be used in a few more places in a later refactor. But, I'm fine removing it too.

core/src/context.rs
```diff
@@ -425,29 +449,42 @@ where
...
+        let wasi_env = {
+            // Get all the variables that this element has read access to
+            let vars_to_read: Vec<(String, String)> = self
+                .get_vars_to_read(from, output_format)?
+                .into_iter()
+                .filter_map(|(name, ty)| {
+                    self.state
+                        .variables
+                        .get(&name)
+                        .filter(|value| value.get_type() == ty)
```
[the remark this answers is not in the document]
adelhult: We talked about this in person so I'll mark it as resolved. I think it should be pretty reasonable decision. But, yeah it can be pretty confusing how we should deal about potential type mismatches. I guess the risk of doing it this ways is that users may use the wrong type and not get a error about doing so. But in most cases where a variable is "required" the package can create a warning or error about "variable ____ is not defined"

core/src/std_packages.rs
```diff
@@ -174,41 +260,125 @@ pub fn native_block_content<T, U>(
...
+    let name = args.get("name").unwrap().as_str().unwrap();
+
+    match ctx.state.variables.get(name) {
+        Some(value @ variables::Value::List(_)) => Ok(text_element(value.to_string(), id.clone())),
+        Some(value) => Err(CoreError::TypeMismatch {
+            name: name.to_string(),
+            expected_type: VarType::List,
+            present_type: value.get_type(),
+        }),
+        None => Ok(Element::Compound(vec![])), // TODO: give a error or warning
```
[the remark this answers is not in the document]
adelhult: I did as we talked about and added a `[list-read foo default="some fallback string"]` maybe not the nicest solution but def. better then what I had before : )

packages/code/src/main.rs
CMDJojo: This issue still remain

packages/code/src/main.rs
CMDJojo: This issue still remain

core/src/variables.rs
```diff
@@ -1,28 +1,180 @@
...
+    }
+
+    /// Clears the variable store
+    pub fn clear(&mut self) {
+        self.0.clear()
+    }
+
+    /// Ensure that a valid name is used.
+    /// Only ASCII alphanumerics characters and "_" is allowed. The name may also not start with a number.
+    fn validate_name(&self, name: &str) -> Result<(), CoreError> {
```
CMDJojo, line 24: This issue doesn't seem to be resolved (not having it as an instance function, plus maybe renaming)

core/src/std_packages.rs
```diff
@@ -176,44 +289,172 @@ pub fn native_block_content<T, U>(
...
+        Some(value) => Err(CoreError::TypeMismatch {
+            name: name.to_string(),
+            expected_type: VarType::Set,
+            present_type: value.get_type(),
+        }),
+        None => {
+            let default = args.get("default").map(ArgValue::as_str).flatten().unwrap();
+            if default == "<log error>" {
+                // No default was provided as fallback, let's log a error
+                log_var_error(ctx, name, "set-read", format);
```
CMDJojo, line 432: This should generate an `[error]` node to facilitate inline error reporting and keep our convention

core/src/std_packages.rs
```diff
@@ -176,44 +289,172 @@ pub fn native_block_content<T, U>(
...
+        Some(value) => Err(CoreError::TypeMismatch {
+            name: name.to_string(),
+            expected_type: VarType::List,
+            present_type: value.get_type(),
+        }),
+        None => {
+            let default = args.get("default").map(ArgValue::as_str).flatten().unwrap();
+            if default == "<log error>" {
+                // No default was provided as fallback, let's log a error
+                log_var_error(ctx, name, "list-read", format);
```
CMDJojo, line 387: This should generate an `[error]` node to facilitate inline error reporting and keep our convention

core/src/std_packages.rs
```diff
@@ -176,44 +289,172 @@ pub fn native_block_content<T, U>(
...
-/// Example function for pushing a string to a list
-pub fn native_push_list<T, U>(
+/// Helper function to create text elements
+fn text_element(contents: String, id: GranularId) -> Element {
+    Element::Module {
+        name: "__text".to_string(),
+        args: Default::default(),
+        body: contents,
+        inline: true,
+        id: id,
```
CMDJojo, line 299: This can be just `id`:

```suggestion
        id,
```

core/src/std_packages.rs
```diff
@@ -176,44 +289,172 @@ pub fn native_block_content<T, U>(
...
+        Some(value) => Err(CoreError::TypeMismatch {
+            name: name.to_string(),
+            expected_type: VarType::Constant,
+            present_type: value.get_type(),
+        }),
+        None => {
+            let default = args.get("default").map(ArgValue::as_str).flatten().unwrap();
+            if default == "<log error>" {
+                // No default was provided as fallback, let's log a error
+                log_var_error(ctx, name, "const-read", format);
```
CMDJojo, line 342: This should generate an `[error]` node to facilitate inline error reporting and keep our convention

CMDJojo (changes requested):
Nice job with the PR and thank you for the changes! However, it seems like some fixes I thought was important didn't quite make it in the code. Please change accordingly

CMDJojo (approved):
Thank you for the changes. Ready to merge!

Issue #128, opened 2023-03-02
Add `any` as special output format


CMDJojo:
**Is your feature request related to a problem? Please describe.**
When making concrete modules that gives one visible output-dependent construction as output, like `table` which makes tables and `math` which renders LaTeX math, it is important to be able to specify the supported output formats. This allows other packages to provide the same module for other output formats, which is very nice. However, for modules which don't give one concrete behaviour, think an `if` module that may ignore its output or may output in ModMark and re-parse it, isn't at all important what the output format is.

**Describe the solution you'd like**
I would need to add a special output format called `any` which says that the module may be used targeting any language. To avoid collisions, we would possibly want that if a module named `x` tagets `any` format, no other modules targeting any formats may be named `x`. This would avoid someone thinking they would run one module while another is actually being run.

**Describe alternatives you've considered**
We could allow one module targeting `any` and one targeting a specific language, like `HTML`, and if `HTML` is selected, the one targeting that will be preferred. This would allow "overriding" the module for some output formats, but it would probably not be that useful and may lead to confusion.

We could also not implement the `any` output format, meaning that the `if` module must specify all possible output types. This would in practice probably lead to that if we make one, it would only support `HTML` and `LaTeX`, meaning that if someone makes a new language plugin for, let's say `Markdown`, and want to use the `if` module, they must implement their own. It probably doesn't require an explanation that multiple implementations of the same thing will lead to poorer performance and unsynced behaviour when the modules may not act the same depending on the output format.

PR #232, opened 2023-04-15
Add any as special output format

Forsinge:
Resolves: #128 

TODO: 
- [x] update existing comments and comment some of the new code
- [x] test collision behaviour

To try it out you can use the `[include]` module, which I've changed to use "any" as output format. 

Collisions seem to work as intended. I tested this by adding a `[hello] -> any` transform and then importing hello from the catalog.

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -97,6 +97,7 @@ impl<T, U> Debug for Context<T, U> {
 pub enum TransformVariant {
     Native((Transform, Package)),
     External(HashMap<OutputFormat, (Transform, Package)>),
+    Any((Transform, Package)),
 }
 
 impl TransformVariant {
@@ -107,6 +108,7 @@ impl TransformVariant {
         match self {
             TransformVariant::Native(t) => Some(t),
             TransformVariant::External(map) => map.get(format),
+            TransformVariant::Any(t) => Some(t),
         }
     }
 
@@ -122,6 +124,7 @@ impl TransformVariant {
             TransformVariant::External(map) => {
                 map.insert(format, entry);
             }
+            TransformVariant::Any(_) => {}
         }
     }
 }
@@ -210,7 +213,7 @@ impl<T, U> Context<T, U> {
     ) -> Result<Vec<(Variable, VarAccess)>, CoreError> {
         // Find the transform to get knowledge about variables
         let Some((transform, package)) = self.package_store.lock().unwrap().find_transform(name, format) else {
-            return Err(MissingTransform(name.to_string(), format.0.clone()));
+            return Err(MissingTransform(name.to_string(), format.to_string()));
         };
 
         if !transform.has_argument_dependent_variable() {
@@ -396,7 +399,7 @@ where
                     named: Some({
                         let mut map = HashMap::new();
                         map.insert("source".to_string(), name.to_string());
-                        map.insert("target".to_string(), output_format.0.to_string());
+                        map.insert("target".to_string(), output_format.to_string());
                         // these two ifs can't be joined, unfortunately, or it won't run on stable
                         if self.state.verbose_errors {
                             map.insert("input".to_string(), data.to_string());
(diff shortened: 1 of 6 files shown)


axoxelol1 (approved):
Seems to work as expected, nice work! Cool to see a custom deserializer as well. Will have a look at #233 once this is merged.

CMDJojo:
I have also looked a bit at the code, it looks great! However, I do have one question; why would we need a custom any `OutputFormat`? Is it just to detect the presence of "any" in the manifest? Since it doesn't really get stored in the map, since the "any" is encoded as another variant, right? Do you think there is a good reason to have "any" as a special variant rather than checking if the OutputFormat string is "any"?

CMDJojo:
Also, about the collision behavior; what happens if a package declares one "normal" output format and "any", like `to: [html, any]`? Do we have a custom error message for that or is it just handled like a normal collision? (Nothing I think is very important, just out of curiosity). I am unable to test the code at this moment, if you want a review from me I can do that tmrw but feel free to merge since Axel already approved!

Forsinge:
Thanks for the review and comments. 

The custom deserializer is definitely not a necessity. As the code is structured right now we could simply perform the string comparison in `insert_transforms`. My thinking was that it is better practice to contain the string manipulations and comparisons to `OutputFormat`, and that it is less error prone to do so if we change the code later on.

I actually forgot to consider the collision scenarion in your second comment. I have yet to test it but it should result in the same error as collisions between packages. Now I'm wondering though, should we let "any" take precedence in this case, set that as `OutputFormat` for that package and ignore other specified formats?

axoxelol1:
Hm the collision is probably a nonproblem but we should still define the expected behavior and test it. I don't think we should let `any` take precedence over other output formats if multiple are specified for a transform. If a package specifies `any` as one of multiple output formats it should probably throw an error since it does not make sense.

CMDJojo:
Thank you for your answers!

Your implementation seems sound, and you are probably right to add an `any` `OutputFormat`. Keep it as-is!

I do not think `any` should take precedence, if any package specifies both `HTML` and `any`, there is probably an error in the module (it is simply just not intended to be done that way). If you want, you can add a check to `Package::verify` that checks for such errors, and in that way it will give an error as soon as the package is loaded rather than when the transform is loaded (it would be possible to load an erroneous package containing `any` and `html` and hide that transform, thus not yielding any error). I think that would be preferable, but it isn't a must.

But good job with the implementation!

core/src/lib.rs
```diff
@@ -59,39 +61,106 @@ impl AccessPolicy for DefaultAccessManager {
...
+        }
     }
 }
 
 impl FromStr for OutputFormat {
     type Err = ();
 
     fn from_str(s: &str) -> Result<Self, Self::Err> {
-        Ok(OutputFormat::new(s))
+        if s == "any" {
```
CMDJojo, line 159: This should probably be an case-insensitive check

core/src/context.rs
```diff
@@ -92,26 +92,28 @@ impl<T, U> Debug for Context<T, U> {
...
-    /// transform, that is returned regardless of the output format, but if it is external, the
-    /// map is searched to find the appropriate transform
+    /// This function finds the transform to an output format. If this if of the `External` variant,
+    /// the map is searched to find the appropriate transform. If this is of the `Native` or
+    /// `ExternalAny` variant, the transform is returned regardless of the output format.
     pub(crate) fn find_transform_to(&self, format: &OutputFormat) -> Option<&(Transform, Package)> {
         match self {
             TransformVariant::Native(t) => Some(t),
             TransformVariant::External(map) => map.get(format),
+            TransformVariant::ExternalAny(t) => Some(t),
```
CMDJojo, line 111: These cases can be collapsed to one;
```rs
            TransformVariant::External(map) => map.get(format),
            TransformVariant::ExternalAny(t) | TransformVariant::Native(t) => Some(t),
```

CMDJojo (commented):
Great job! I have only looked through the code briefly, I don't think I need to approve it since Axel already did! I found two nit-picks, but feel free to skip them if you want

PR #233, opened 2023-04-16
Add `[if]` module and flow package

Forsinge:
Resolves: #212 

Note that this PR depends on #232 and assumes that it has been merged


CMDJojo:
The `if` transform is super cool and it is very nice to get conditional compilation in this way! I have myself thought of making such a transform after adding `any`, and when I thought about it I though about having conditional based on output format and variables. However, due to how we structured variables, I think we would need one `if`, one `if-list`, one `if-const` etc, so I think that is out of scope of this PR. I think this should be added as-is, and I (or any other) may add variable-dependent conditionals later on if needed.

Two things I noted:
* I think comparing the output format should be case-insensitive, so `[if format=HTML]` would work
* Maybe we would want to allow negative matching, like `[if format=!HTML]` (maybe just prefixing with `!`). I don't know the use cases of this, but it would maybe be nice

Forsinge:
Opened this for review now because #232 has been merged. I changed to case-insenitive comparisons, and I also think the negative matching seems like a neat idea so I added that as well!

CMDJojo (approved):
Looks good to me! Nice to have some conditional compilation! The only thing I think can be improved is maybe add a description to the transform itself, like "Conditionally compile content based on output format", but skip that if you think it is unnecessary. Good job!

CMDJojo:
I am running a manual teardown of the preview of this branch since it didn't seem to happen automatically for some reason

Issue #237, opened 2023-04-17
Add package for plotting graphs


Forsinge:
**Is your feature request related to a problem? Please describe.**
A package to create graphs is a useful addition, especially when it comes to report writing. The package may be more suited to be placed in the ModMark catalog, but for now it is simpler to include it in std.

Two modules that could be included in the package are:
- a module to plot functions
- a module to plot a set of values/coordinates

PR #238, opened 2023-04-18
Add plot package

Forsinge:
Resolves: #237 

This PR adds the plot package, which includes the `[plot]` and `[plot-list]` modules.

Current features: 
- plot arbitrary functions of one variable (now you can plot several functions in the same graph!)
- plot a list of points
- specify axis ranges
- sqrt, log, mod and common trigonometric functions are supported inside functions
- store the resulting SVG as a file
- decent error handling (no custom error type though), verification that the provided function is valid
- figure, caption and label
- set line width used in plot

Known limitations/bugs:
- ~~only x,y,z are supported as a variables (and they are treated as the same variable)~~ still restricted to single characters
- ~~plot cuts off early for certain functions~~
- ~~unexpected jumps close to y-bounds for certain functions~~
- ~~function args such as "2x" do not work (example: "sin(2x)")~~ sin(2*x) works now, but not sin(2x)
- ~~squiggly plot (don't know if this is just a result of floating point arithmetic)~~ mostly solved, also added "samples" arg
- ~~suboptimal plots of functions such as 1/x, ideally it should produce two separated lines~~ mostly solved
- ~~there seems to be some issues with operator precedence as well~~

Example usage: 
```
[plot 0 20 0 20 save="myplot.svg"]
4*sin(x)+x
20-x^2
```
```
[plot-list line_width=2 connect=true point_size=4]
0 0
5.3 2.5
0 4
3.2 -0.8
3.2 5
0 0
```


Forsinge:
Opening for review now! The plots are not flawless, but should be good for most functions and axis ranges. There is quite a lot going on in the modules, so I'm open to suggestions on how to improve them. I encourage you to play around with the plots and find problems :)

I've thought a bit about how to add LaTeX support. The best I've come up with is always saving the resulting SVG to a file, and then generating an `[image]` module pointing to the saved file. This approach requires the path to be accessible both from the LaTeX document and the [plot] module itself. Thoughts?

Some changes to note since last commit regarding usage:
-`[plot_list]` has been renamed to `[plot-list]`
- `[plot]` now supports more than one function (separated by newlines)
- There is now a `samples` argument  to set the "granularity" of the plot. Generally, fewer samples means smoother plot.
- There are now separate arguments for each axis boundary, instead of `x_range` and `y_range`. This means you can write `[plot -10 10 0 20]`
- You can no longer set the color of the plot. This became messy to implement when supporting plots for several functions.

adelhult:
Nice work! Will take a look : )

> I've thought a bit about how to add LaTeX support. The best I've come up with is always saving the resulting SVG to a file, and then generating an `[image]` module pointing to the saved file. This approach requires the path to be accessible both from the LaTeX document and the [plot] module itself. Thoughts?

Seems like a good idea! I've been thinking about a similar problem for the chalmers-thesis template. The problem with saving a file is that we don't have any control of the output path and we don't know if that will link up with the assets folder correctly. I still think it's good approach and not a major problem. Another possibility that has it's own trade-offs is embedding the svg directly in the .tex file. I'm actually looking at that right now :sweat_smile: maybe the [`\filecontents`](http://latexref.xyz/filecontents.html) command can be of help. 

image

packages/plot/src/utils.rs
```diff
@@ -0,0 +1,402 @@
...
+
+pub fn print_svg_html(svg: String, ctx: &PlotContext) {
+    let encoded: String = general_purpose::STANDARD_NO_PAD.encode(&svg);
+    let src = format!("data:image/svg+xml;base64,{encoded}");
+    let percentage = (ctx.svg_info.width * 100.0).round() as i32;
+    let style = format!("style=\"width:{percentage}%\"");
+    let img_str = format!("<img src=\"{src}\" {style} ");
+
+    let mut v = vec![];
+    v.push(raw!("<figure>\n"));
```
adelhult, line 229: Since we now (as far as I know?) just support string literals for raw modules we maybe be able to remove the use of the `raw!` macro in favour of using the string : )

packages/plot/src/plots.rs
```diff
@@ -0,0 +1,194 @@
...
+use crate::utils::*;
+
+type Points = Vec<(f64, f64)>;
+type Plot = Vec<Points>;
+
+fn get_function_plots(ctx: &mut PlotContext) -> Result<Vec<Plot>, String> {
+    let mut plots = Vec::new();
+
+    for function in ctx.data.split('\n') {
+        let mut x_iter = ctx.rm.get_point_range('x')?;
```
adelhult, line 14: I encountered an issue were a trailing newline or whitespace character would cause the module to fail. Maybe add something along these lines. Note: I've not tested my suggestion, so it might not compile. But you get the idea :).
```suggestion
        // skip lines only containing whitespace
        if function.chars().all(char::is_whitespace) {
            continue;
        }

        let mut x_iter = ctx.rm.get_point_range('x')?;
```

packages/plot/src/eval.rs
```diff
@@ -0,0 +1,253 @@
...
+fn value_to_float(v: &Value) -> Option<f64> {
+    match v {
+        Value::Float(f) => Some(*f),
+        Value::Int(i) => Some(*i as f64),
+        _ => None,
+    }
+}
+
+// Create a new function context with functions sin, cos, tan, sqrt, log, mod
+pub fn new_function_context() -> HashMapContext {
```
adelhult, line 17: Nice! Maybe add arcsin, arctan etc. and some constants like PI.

packages/plot/src/eval.rs
```diff
@@ -0,0 +1,253 @@
...
+            if !vars.contains(&s) {
+                vars.push(s);
+            }
+        }
+    }
+    vars
+}
+
+// Verify all the functions provided by the user
+pub fn verify_functions(ctx: &mut PlotContext, var_count: usize) -> Result<(), String> {
```
adelhult, line 131: Great that you do this check for better error messages! I've not looked close enough to know for sure, but could we not look at the `evalexpr::error::EvalexprError` instead? Like the FunctionIdentifierNotFound variant

adelhult (commented):
Great work, it was really fun to play around with : )

I left a couple of suggestions to both the code and API!

packages/plot/src/utils.rs
```diff
@@ -0,0 +1,402 @@
...
+                            "name": "save",
+                            "description": "The name of the SVG file that is saved. No file is saved if this argument is left empty.",
+                            "default": "",
+                        },
+                        {
+                            "name": "line_width",
+                            "description": "The width of the line that is used in the plot.",
+                            "type": "uint",
+                            "default": 1,
+                        },
```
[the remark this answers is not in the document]
Forsinge: Decided to let the user specify hex codes for these, same for [plot-list].

adelhult (approved):

PR #239, opened 2023-04-19
Handle "black boxes" and ensure that it is not possible to lie to the scheduler

adelhult:
We want to inform the scheduler about transforms that produces arbitrary content (that may access any variable) so these modules can be evaluated first. I'm calling the property `black-box` for now tto avoid to much bikeshedding, but I agree with @CMDJojo that we should probably use another name. I think his suggestions are good, maybe `arbitrary_content` or `unrestricted_var_access` could work too? I don't really have a strong opinion.

We also want to make sure that packages can't lie about what variables they access and break the scheduling. We will likely have some sort of object that describes the variable privileges that the module has and let new elements that are spawned by that element inherit the same privileges and then check for errors later when evaluating those new elements)  

A simple test case for the `black-box` property would be:
```
[inline_content]{
    [list-push foo] 1
}


[list-push foo] 2

[list-read foo]
```

Testing the privilege system could look something like this perhaps:
```
[id_functions_without_var_access]
[inline_content] inline content requires full access since it may contain anything
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,15 +6,16 @@ Welcome to the main ModMark repo! This is the home of the ModMark modular markup
 
 To invoke the ModMark compiler, you can run
 
-> cargo run -- in.mdm out.html
+> cargo run -- compile in.mdm out.html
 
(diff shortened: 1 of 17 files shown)


adelhult:
Looks like @CMDJojo already implemented this so I'll close the PR : )

Issue #225, opened 2023-04-14
Support black boxes/generating transforms/parsing transforms


CMDJojo:
Quote from #224: 

> "Black boxes" are not implemented in this PR either. A "black box"/"generating transform"/"parsing transform" (name not decided) is a transform that by itself doesn't require access to any variables but may generate content that does. One example is that `[table]`, which doesn't require access to any variables, may generate a `[push-list]` module, which does. In this current implementation, there is nothing saying that the `[table]` transform must be run before any `[read-list]` transforms, meaning that it is arbitrary which gets run first. To get a consistent behaviour of this, there must be some notion of "these things must be evaluated with a high priority, because they may generate things that have other, currently unknown, dependencies". Since terminology isn't decided, yet alone how it should be handled, I think we should hold back with that a bit and go ahead with this PR to ease the development of a variable store. The scheduler will be easy to adapt to take this into account, and I will promptly create a new issue regarding these "black boxes".

As the quote describes, we need to have a notion of transforms that may re-parse its content that may possibly have other dependencies than the transform itself. This is to make sure that transforms such as "table", "center", "row" etc executes before "toc", "document" and so on.

## Naming
Here are some possible names:
* Black box transform (or just Black box)
* Generating transform
* Parsing transform
* Re-parsing transform
* Expanding transform
* Recursive transform

I think Black box transform looks cool, but I would argue that "Generating transform" is the name that conveys the meaning the best. It is simply a transform which may generate content that does more than just insert raw nodes, image tags etc. However, developers may think of it like "Doesn't every transform generate something", which would be bad. "Recursive transform" is probably not very good since recursive functions generally call themselves, while this doesn't necessarily. "Parsing transform" may sound great, but there are ways of generating unknown dependencies without actually re*parsing*. I am unsure what to call this, please comment suggestions below!

## Implementation
We might add an (optional) boolean field to the transform in the manifest, so that `[table]` can be defined like:
```json
{
    "from": "table",
    "to": ["html", "latex"],
    "arguments": [
        {"name": "caption", "default": "", "description": "The caption for the table"},
        {"name": "label", "default":"", "description": "The label to use for the table, to be able to refer to it from the document"},
        {"name": "header", "default": "none", "type": ["none", "bold"], "description": "Style to apply to heading, none/bold"},
        {"name": "alignment", "default": "left", "description": "Horizontal alignment in cells, left/center/right or l/c/r for each column"},
[...]

PR #240, opened 2023-04-19
GH-225: Implement `unknown_content` for transforms

CMDJojo:
This PR resolves GH-225 by adding an optional field `unknown-content` to the manifest for a transform. Setting this field to true ensures that that module is executed before any module which has that field set to false.

Take this as an example:
```
[push-list abc] a

[inline_content]
[push-list abc] b

[read-list abc]
```

We have three modules, `push-list`, `inline_content` and `read-list`. We know that the scheduler will ensure `push-list` occurs before `read-list`, but the scheduler doesn't know anything about `inline_content` so it may schedule it whenever it sees fit. If it schedules it before evaluating `read-list`, the dependency will be picked up as expected but if it schedules it later, `read-list` would already have been executed.

This PR changes it so that `inline_content` in this case has *unknown content*, meaning that it will be executed before anything with known content. Thus, this hazard is no more.

This PR is ready for review immediately

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -191,6 +191,11 @@ where
     }
 }
 
+pub struct Dependencies {
+    pub var_accesses: Vec<(Variable, VarAccess)>,
+    pub has_unknown_content: bool,
+}
+
 impl<T, U> Context<T, U> {
     /// This function returns the "variable accesses", which is all variables the transform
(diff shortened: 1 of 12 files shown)


adelhult (approved):
Looks good, tested it out for a bit too, I did not encounter any issues!

Issue #246, opened 2023-04-20
Disable verbose errors by default


adelhult:
I'm building two larger document templates (acm and chalmers-thesis) where I want to log errors in the __document transform when keys are missing the warning output gets borderline unreadable due to the input field including the entire document one time for each warning. 

I think our error messages would end up looking a bit better if we disable verbose errors by default at least for now until we have improved how they are displayed on the web and in the browser.

PR #247, opened 2023-04-20
Disable verbose errors and add --verbose CLI flag 

adelhult:
Resolves GH-246. I added a `--verbose` flag as @CMDJojo suggested! : )

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -35,6 +35,8 @@ pub struct Context<T, U> {
     engine: Engine,
     pub(crate) state: CompilationState,
     pub filesystem: CoreFs<U>,
+    /// If true verbose errors and warnings will be used
+    pub verbose: bool,
     policy: Arc<Mutex<U>>,
 }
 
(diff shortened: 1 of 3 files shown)


CMDJojo:
Hmm, it would be much better to have it configurable from outside. Maybe it would be possible without too much work to take it as an argument to the `eval` function and have a CLI flag (and hard-code it in the playground until we implement a switch there)?

core/src/context.rs
```diff
@@ -58,18 +59,17 @@ pub struct Issue {
...
-        if let Some(input) = &self.input {
-            write!(
-                f,
-                "{} -> {}: {}, input: {}",
-                self.source, self.target, self.description, input
-            )
-        } else {
-            write!(
+        match self.input.as_ref().map(String::as_str) {
+            Some("input") | None => write!(
```
CMDJojo, line 63: What does this matching against `Some("input")` do? When is the `input` argument equals to `input`? I don't think you did something wrong, I'm just curious

CMDJojo (approved):
Good job! Will be nice to be able to skip verbose errors, they are not always very useful

core/src/context.rs
adelhult: Oops! I think it actually is a type and should say something along the lines of "<unknown>" and not "input". Good catch! I'll fix that before I rebase

Issue #248, opened 2023-04-20
Add LaTeX imports using variables


CMDJojo:
Now when variable dependencies were added to the manifest as a part of #224 and an interface to variables was added as a part of #231, it makes sense to use the variables. One place where this can be very useful is to use `set`s to manage imports in LaTeX (`\usepackage`s). It would be appropriate to have a variable `imports` which includes the entire `\usepackage` statements, since we can't just store the names due to some of them needing configurations and such.

PR #249, opened 2023-04-20
Use set for LaTeX imports

CMDJojo:
This PR resolves GH-248 by adding an `imports` set to store what imports are needed. Entries are added to this set via the `[set-add]` module, and a macro was made to simplify the use of this:
```rust
macro_rules! import {
    ($e:expr) => {json!({"name": "set-add", "arguments": {"name": "imports"}, "data": $e})}
}
```
The macro is similar to the `raw!` macro, but generates a `set-add` instead.

The transforms that relies on imports have been updated to push their imports to the set. This applies to all packages.

All previously hardcoded packages are now removed from the prelude in `__document`, and when transforming that, the imports are read and parsed as a JSON array (or as an empty array if the key doesn't exist), and a prelude is created with those imports.

This PR is ready for review immediately.

### DIFF
--- a/packages/files/src/main.rs
+++ b/packages/files/src/main.rs
@@ -1,10 +1,11 @@
-use base64::{engine::general_purpose, Engine as _};
-use serde_json::{json, Value};
 use std::env;
 use std::ffi::OsStr;
 use std::fs;
 use std::io::{self, Read};
 use std::path::Path;
+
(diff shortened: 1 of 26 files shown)


adelhult (commented):
Nice work!

adelhult (approved):

PR #251, opened 2023-04-20
Variable based conditional compilation

### LINKED ISSUE #250: Variable based conditional compilation
As a part of #233, the `[if]` module was added to compile content conditionally based on output format. This is great, and would allow us to, for example, include some HTML widget when compiling to HTML but not when compiling to LaTeX.

Now when variables have been added, it would be nice to have similar modules that compile conditionally based on variables. Based on the manifest variable structure, it would probably be best to have multiple modules, like `[if-const]`, `[if-list]` etc for the different data types.

CMDJojo:
This PR resolves #250 by adding three new modules, `[if-const]`, `[if-list]` and `[if-set]`. They all have he same general syntax, `[if-<type> <variable> [check] [value] [case-sensitive]]` where `<type>` is the type (`const`/`list`/`set`), `<variable>` is the name of the variable to check, `[check]` is the way to check the variable (like the operator) and `[value]` is the value to check against. `[case-sensitive]` is set to `case-sensitive` by default and affects all string comparisons.

`[if-const]` has these checks:
* **defined** succeeds if the constant is defined (default)
* **undefined** succeeds if the constant is undefined
* **equals** succeeds if the constant value equals *value*
* ~~differs-to~~ **differs-from** succeeds if the constant value differs from *value*

Here are some examples:
```
[const-declare foo] abc
[const-declare bar] def

[if-const foo]
Foo is defined!

[if-const foo equals abc]
Foo is 'abc'!

[if-const bar differs-from abc]
Bar differs to 'abc'

[if-const bar equals abc]
This isn't going to be compiled

[if-const xyz undefined]
xyz is undefined
```

`[if-list]` and `[if-set]` has the same checks, and for all intents and purposes the set is just treated as a list. These are the checks available:
* **does-not-contain** succeeds if *value* isn't in the list/set
* **contains** succeed if *value* is in the list/set
* **contains-all** succeed if all comma-separated values is in the list/set
* **contains-any** succeed if any of the comma-separated values are in the list/set
* **contains-none** succeed if none of the comma-separated values is in the list/set
* **non-empty** succeed if the list/set is non-empty (and thus defined, this is default)
* **empty-or-undefined** succeed if the list/set is either empty or undefined

Since it was decided beforehand that it is important to have a distinction between a list/set being undefined and being empty, I am using the term `empty-or-undefined` to signal any of these states.

Here are some examples:
```
[list-push authors] Jonathan
[list-push authors] Eli
[list-push authors] Hugo
[list-push authors] Gustav
[list-push authors] Axel
[list-push authors] Carl

[if-list authors contains-all jonathan,eli,hugo]
This is a case-sensitive search by default so this won't be compiled

[if-list authors contains-all jonathan,eli,hugo case-insensitive]
We have set the authors field correctly!

[if-list authors contains-none Isak,Lars,Fredrik]
There are no authors named Isak, Lars or Fredrik
```

One larger example document:
```
[const-decl theme] dark

[if-const theme equals dark]
This is **//dark theme//**

[if-const theme undefined]
[warning]
Please define theme

[list-push authors] Jonathan
[list-push authors] Eli
[list-push authors] Hugo
[list-push authors] Gustav
[list-push authors] Axel
[list-push authors] Carl

[if-list authors contains Jonathan]
Hi, im Jonathan

[if-list authors contains-all jonathan,hugo,eli]
This is a case-sensitive search by default so this won't be compiled

[if-list authors contains-all jonathan,hugo,eli case-insensitive]
We have set the authors field correctly!

[if-list authors contains-none Isak,Lars,Fredrik]
There are no authors named Isak, Lars or Fredrik

[if-list authors contains-any Isak,Lars,Fredrik,Eli]
(At least) one of Isak, Lars, Fredrik or Eli are mentioned as authors

[if-list foo empty-or-undefined]
List foo is empty!
```

### DIFF
--- a/packages/flow/src/main.rs
+++ b/packages/flow/src/main.rs
@@ -194,7 +194,7 @@ fn transform_if_const(input: Value) {
     } else if let Ok(var_val) = env::var(var_name) {
         let case_sensitive = input["arguments"]["case-sensitive"].as_str().unwrap() == "true";
         let var_val = if case_sensitive {
-            var_val.to_string()
+            var_val
         } else {
             var_val.to_lowercase()
         };
@@ -237,27 +237,36 @@ fn transform_if_collection(input: Value, is_set: bool) {
     }
 
     let check = input["arguments"]["check"].as_str().unwrap();
-    let value = input["arguments"]["value"].as_str().unwrap();
+    let value = {
+        let arg = input["arguments"]["value"].as_str().unwrap();
+        if case_sensitive {
+            arg.to_string()
+        } else {
+            arg.to_lowercase()
+        }
+    };
+
+    // We have a Vec<String> and want to check if it contains a &str, but there is no way of doing
+    // that using .contains since it expects a &String (and you can make a &String to a &str but
+    // not the other way around). Because of that, we do .iter().any(|v| v == x) with this macro
+    macro_rules! contains {
+        ($c:expr, $str:expr) => {
+            $c.iter().any(|v| v == $str)
+        };
+    }
 
     let compile = if check == "non-empty" || check == "empty-or-undefined" {
         env_values.is_empty() == (check == "empty-or-undefined")
     } else if check == "contains" || check == "does-not-contain" {
         // cannot use .contains here due to coercion rules
-        env_values.iter().any(|v| v == value) == (check == "contains")
+        contains!(env_values, &value) == (check == "contains")
     } else {
-        let split_values: Vec<_> = if case_sensitive {
-            value
-                .split_terminator(",")
-                .map(ToString::to_string)
-                .collect()
-        } else {
-            value.split_terminator(",").map(str::to_lowercase).collect()
-        };
+        let split_values: Vec<&str> = value.split_terminator(',').collect();
 
         match check {
-            "contains-all" => split_values.iter().all(|v| env_values.contains(v)),
-            "contains-any" => split_values.iter().any(|v| env_values.contains(v)),
-            "contains-none" => split_values.iter().all(|v| !env_values.contains(v)),
+            "contains-all" => split_values.into_iter().all(|v| contains!(env_values, v)),
+            "contains-any" => split_values.into_iter().any(|v| contains!(env_values, v)),
+            "contains-none" => split_values.into_iter().all(|v| !contains!(env_values, v)),
             _ => unreachable!("Invalid check {check}"),
         }
     };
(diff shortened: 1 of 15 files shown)


CMDJojo:
Ps. I know that I could use the enums `true`/`false` for `case-sensitive` rather than `case-sensitive`/`case-insensitive`, but I think `[if mode equals dark case-insensitive]` looks better since you can immediately know what that last argument is doing rather than doing `[if mode equals dark false]`, where it isn't clear at all what the last argument refers to (if you don't specify the key)

packages/flow/src/main.rs
```diff
@@ -41,6 +41,126 @@ fn manifest() {
...
+                        {
+                            "name": "check",
+                            "description": "Specifies the check to do to the value. \
+                            equals: compile the content if the variable equals the given value. \
+                            differs-to: compile the content if the variable differs from the given value. \
+                            defined: compile the content if the variable is defined. \
+                            undefined: compile the content if the variable is undefined.
+                            ",
+                            "default": "defined",
+                            "type": ["equals", "differs-to", "defined", "undefined"]
```
adelhult, line 64:
```suggestion
                            "type": ["equals", "differs-from", "defined", "undefined"]
```

packages/flow/src/main.rs
```diff
@@ -41,6 +41,126 @@ fn manifest() {
...
+                        {
+                            "name": "check",
+                            "description": "Specifies the check to do to the value. \
+                            does-not-contain: compile the content if the set does not contain the given string. \
+                            contains: compile the content if the set contains the given string. \
+                            contains-all: compile the content if the set contains all of the comma-separated strings. \
+                            contains-any: compile the content if the set contains any of the comma-separated strings. \
+                            contains-none: compile the content if the set contains none of the comma-separated strings. \
+                            non-empty: compile the content if the set is defined and has at least one string. \
+                            empty-or-undefined: compile the content if the set is undefined or is empty
```
adelhult, line 102: as an alternative we could have a `empty` and `undefined` option that both do the same thing if you prefer that.

adelhult (approved):
Looks good! Left two small comments (feel free to ignore the one regarding empty-or-undefined)

packages/flow/src/main.rs
CMDJojo: Fixed!

packages/flow/src/main.rs
CMDJojo: As discussed in private, since a list can't be empty (just undefined), `empty` either has to always fail (which is unexpected) or lie. Additionally it may be confusing to have two options doing the same thing. So I will skip this one

Issue #254, opened 2023-04-24
Add bibliography package


CMDJojo:
It would be nice to have a bibliography package which allows us to import a BibLaTeX file (or similar) and use that for citations in the text and for making a bibliography at the end of the document.

PR #255, opened 2023-04-24
GH-254: Add bibliography package

CMDJojo:
This PR resolves GH-254 by adding a bibliography package. The package generates bibliographies, surprise surprise.

I think I am more or less done with this package, but it may have some rough edges, so please come with comments if you have any. There are two new modules intended for use; `[cite [note]] <name>`  will cite a source from the bibliography by its name, optionally allowing a note such as page numbers, and `[bibliography [style] [file]] <database>` which imports a BibTeX or Hayagravia Yaml file and uses that to create a bibliography and resolve previous citations. The BibLaTeX (or Yaml) content may be in a file (where the filename is passed as argument), or inserted as the body of the module.

Example document:
```
## Chapter 1

John MacFarlane [cite "p. 22"] djot released a product
called djot, after releasing Commonmark [cite] Commonmark in 2021.
AsciiDoc by Eclipse Foundation [cite] AsciiDoc is also intersting.

## Bibliography

[bibliography IEEE]<<
@online{djot,
title = {djot and the books that actually works bla bla it is nice. This is nice},
author = {John MacFarlane},
url = {https://djot.net},
urldate = {2023-01-23}
}

@online{AsciiDoc,
author = {{Eclipse Foundation}},
title = {{AsciiDoc}},
url = {http://asciidoc.org/},
urldate = {2023-01-23}
}

@online{Commonmark,
author = {John MacFarlane},
title = {CommonMark Spec},
year = {2021},
month = {6},
url = {https://spec.commonmark.org/0.30/},
urldate = {2023-03-21}
}
>>
```

### DIFF
--- a/packages/bibliography/src/main.rs
+++ b/packages/bibliography/src/main.rs
@@ -1,12 +1,12 @@
+use std::{env, fs};
 use std::borrow::Cow;
 use std::io::{self, Read};
-use std::{env, fs};
 
+use hayagriva::Entry;
 use hayagriva::style::{
     Apa, BibliographyStyle, ChicagoAuthorDate, Citation, CitationStyle, Database, DisplayString,
     Formatting, Ieee, Mla, Numerical,
 };
-use hayagriva::Entry;
 use serde::{Deserialize, Serialize};
 use serde_json::{json, Value};
 
@@ -307,7 +307,7 @@ fn read_bibliography(input: &Value) -> Option<Vec<Entry>> {
         return None;
     };
 
-    let bibliography = hayagriva::io::from_biblatex_str(&input_string)
+    let bibliography = hayagriva::io::from_biblatex_str(&without_latex_comments(&input_string))
         .or_else(|e1| hayagriva::io::from_yaml_str(&input_string).map_err(|e2| (e1, e2)));
 
     match bibliography {
@@ -355,4 +355,11 @@ fn display(string: &DisplayString) -> Vec<Value> {
     }
     values.push(text!(&string.value[last_unformatted..]));
     values
+}
+
+fn without_latex_comments(string: &str) -> String {
+    string.lines()
+        .filter(|line| !line.trim_start().starts_with('%'))
+        .collect::<Vec<_>>()
+        .join("\n")
 }
--- a/packages/bibliography/src/main.rs
+++ b/packages/bibliography/src/main.rs
@@ -27,6 +27,16 @@ macro_rules! module {
     ($name:expr, $data:expr $(,$($args:tt)*)?) => {json!({"name": $name $(,"arguments":$($args)*)*, "data": $data})}
 }
 
+macro_rules! raw {
+    ($content:expr) => {
+        module!("raw", $content)
+    };
+}
+
+macro_rules! import {
+    ($package:expr) => {module!("set-add", $package, {"name": "imports"})}
+}
+
 macro_rules! push_citation {
     ($e:expr) => {module!("list-push", $e, {"name": "inline_citations"})}
 }
@@ -171,7 +181,7 @@ fn transform_cite_label(_to: &str, input: &Value) {
     }
 }
 
-fn transform_bibliography(_to: &str, input: &Value) {
+fn transform_bibliography(to: &str, input: &Value) {
     let Some(bibliography) = read_bibliography(input) else { return; };
 
     // This is the citations (i.e [cite]s) used in the text
@@ -192,7 +202,12 @@ fn transform_bibliography(_to: &str, input: &Value) {
     let mut output: Vec<Value> = vec![];
     let mut used_keys: HashSet<&str> = HashSet::new();
 
-    let unused_keys_is_visible = input["arguments"]["unused-entries"].as_str().unwrap() == "visible";
+    if to == "latex" {
+        output.push(import!(r"\usepackage[hidelinks]{hyperref}"));
+    }
+
+    let unused_keys_is_visible =
+        input["arguments"]["unused-entries"].as_str().unwrap() == "visible";
 
     for citation_str in &citations {
         // We parse the citation (which is a JSON obj)
@@ -231,16 +246,30 @@ fn transform_bibliography(_to: &str, input: &Value) {
         // From this, we construct our citation string (what the [cite] should be turned into).
         // We pass this as JSON format to let Hayagravia apply formatting if needed (which it
         // doesn't, in this version at least). We encase it in [] for IEE and () for others
-        let json_citation = if style_arg == "IEEE" {
+        let mut json_citation_content = if style_arg == "IEEE" {
             let mut vec = vec![text!("[")];
             vec.append(&mut display(&database_citation.display));
             vec.push(text!("]"));
-            Value::Array(vec)
+            vec
         } else {
             let mut vec = vec![text!("(")];
             vec.append(&mut display(&database_citation.display));
             vec.push(text!(")"));
+            vec
+        };
+
+        let json_citation = if to == "latex" {
+            let mut vec = vec![raw!(format!(r"\hyperlink{{bibentry:{}}}{{", entry.key()))];
+            vec.append(&mut json_citation_content);
+            vec.push(raw!("}"));
             Value::Array(vec)
+        } else if to == "html" {
+            let mut vec = vec![raw!(format!(r##"<a href="#bibentry:{}">"##, entry.key()))];
+            vec.append(&mut json_citation_content);
+            vec.push(raw!("</a>"));
+            Value::Array(vec)
+        } else {
+            Value::Array(json_citation_content)
         };
 
         // Then, we are adding that citation as the label (that the [cite], now [cite-internal],
@@ -267,7 +296,23 @@ fn transform_bibliography(_to: &str, input: &Value) {
                 output.append(&mut display(prefix));
                 output.push(text!("] "));
             }
+            if to == "latex" {
+                output.push(raw!(format!(
+                    r"\hypertarget{{bibentry:{}}}{{",
+                    entry.entry.key()
+                )));
+            } else if to == "html" {
+                output.push(raw!(format!(
+                    r#"<span id="bibentry:{}">"#,
+                    entry.entry.key()
+                )));
+            }
             output.append(&mut display(&entry.display));
+            if to == "latex" {
+                output.push(raw!("}".to_string()));
+            } else if to == "html" {
+                output.push(raw!("</span>"));
+            }
             output.push(module!("newline", ""));
         }
     }
@@ -314,6 +359,9 @@ fn read_bibliography(input: &Value) -> Option<Vec<Entry>> {
     let input_string: Cow<'_, str> = if filename.is_empty() {
         Cow::Borrowed(input["data"].as_str().unwrap())
     } else if let Ok(content) = fs::read_to_string(filename) {
+        if !input["data"].as_str().unwrap().is_empty() {
+            eprintln!("Bibliography contains body text but it is ignored since the 'file' argument is present");
+        }
         Cow::Owned(content)
     } else {
         eprintln!("Could not read file {filename}");
@@ -371,7 +419,8 @@ fn display(string: &DisplayString) -> Vec<Value> {
 }
 
 fn without_latex_comments(string: &str) -> String {
-    string.lines()
+    string
+        .lines()
         .filter(|line| !line.trim_start().starts_with('%'))
         .collect::<Vec<_>>()
         .join("\n")
--- a/packages/bibliography/src/main.rs
+++ b/packages/bibliography/src/main.rs
@@ -1,0 +1,358 @@
+use std::borrow::Cow;
+use std::io::{self, Read};
+use std::{env, fs};
+
+use hayagriva::style::{
+    Apa, BibliographyStyle, ChicagoAuthorDate, Citation, CitationStyle, Database, DisplayString,
+    Formatting, Ieee, Mla, Numerical,
+};
+use hayagriva::Entry;
+use serde::{Deserialize, Serialize};
+use serde_json::{json, Value};
+
+fn main() {
+    let args: Vec<String> = env::args().skip(1).collect();
+    let action = &args[0];
+    match action.as_str() {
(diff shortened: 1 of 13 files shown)


CMDJojo:
Some technical notes:

[typst/hayagriva](https://github.com/typst/hayagriva) is used for generating the bibliography and citations, allowing different styles such as `IEEE`, `APA` `CMoS` etc. The library allows setting two different styles, `BibliographyStyle` for the bibliography and `CitationStyle` for citations. This implementation sets both of them from the bibliography style, but I am not super sure that the `CitationStyle` is set correctly (but for `IEEE` and `CMoS`), the documentation was hard to follow and there didn't exist one `CitationStyle` for each `BibliographyStyle`.

To get the bibliography and cites correct, the scheduler are leveraged to compile the document in three stages:
1. All `[cite]`s are recorded in order (their content and arguments are stored in a list), but no actual citations are added in the text
2. `[bibliography]` runs, knowing all `[cite]`s and the order they appear in, and it makes a bibliography as well as generates the text for the inline citations
3. At the previous position of the `cite`s, the appropriate inline citation text is fetched (from the output of `[bibliography]`) and inserted into the document

The first step is done by pushing JSON objects to the list `inline_citations`, as well as creating new modules `[cite-internal-do-not-use]` which is scheduled to run later. The second step is done by reading `inline_citations`, creating the actual bibliography and citations, and then for each citation in `inline_citations`, push a string containing that JSON and some JSON to be inserted in the modmark document at the citation place to a set called `inline_citation_labels`. The third step is done by having `[cite-internal-do-not-use]` read `inline_citation_labels`, finding its own label and inserting that JSON there.

For this to be possible, `[cite]` has to have `unknown-content:true` while not declaring `inline_citations:list/push` since that would lead to cyclic dependencies. Additionally, `[bibliography]` has declared `imports:set/push` just to facilitate `[link]` importing `hyperref` which it needs for LaTeX output.

Note that all of these modules have output format `any`, since they aren't dependent on the output format. It does, however, require that `__text`, `__italic`, `__bold`, `link` and `newline` exist for the target language.

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+    match action.as_str() {
+        "manifest" => manifest(),
+        "transform" => transform(&args[1], &args[2]),
+        other => {
+            eprintln!("Invalid action {other}");
+        }
+    }
+}
+
+macro_rules! module {
```
adelhult, line 25: Nice to see the macros, it gets a lot cleaner.

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+                {
+                    "from": "cite",
+                    "to": ["any"],
+                    "arguments": [
+                        {"name": "note", "default": "", "description": "Note to show after citation, such as p. 3"}
+                    ],
+                    "description": "Add an inline citation to one of the sources defined in the bibliography. Example: [cite] modmark",
+                    // "variables": {
+                        // We really want this, but we can't since it would give cyclic dependencies
+                        // together with unknown-content. So, we disable unknown-content
```
adelhult, line 67:
```suggestion
                        // together with unknown-content. So, we disable this in favour of just using unknown-content
```

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+                        // "inline_citations": {"type": "list", "access": "push"}
+                    // },
+                    "unknown-content": true // This gives cyclic dependencies if enabled together
+                    // with inline_citations/list-push, see below
```
adelhult, line 71: repeated comment
```suggestion
                    "unknown-content": true
```

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+                },
+                {
+                    "from": "bibliography",
+                    "to": ["any"],
+                    "description": "Inserts a bibliography into the document. \
+                    This module needs to exist for [cite]s to work properly. \
+                    The bibliography may be read from the BibLaTeX or Hayagriva YAML format, \
+                    from either the body of this module or a file passed in as the \"file\" argument.",
+                    "arguments": [
+                        {"name": "style", "default": "IEEE", "type": ["IEEE", "APA", "MLA", "CMoS"], "description": "The style to have the bibliography in"},
```
adelhult, line 90: Maybe just call it Chicago since I think that is more widely known than the abbreviation. 
```suggestion
                        {"name": "style", "default": "IEEE", "type": ["IEEE", "APA", "MLA", "Chicago"], "description": "The style to have the bibliography in"},
```

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+                        {"name": "file", "default": "", "description": "A file containing BibLaTeX or Hayagriva YAML with the bibliography"},
+                        {"name": "visibility", "default": "shown", "type": ["shown", "hidden"], "description": "Whether the bibliography is 'shown' or 'hidden'. \
```
adelhult, line 92: Maybe worth using the same names as the CSS property `visibility` does?
```suggestion
                        {"name": "visibility", "default": "shown", "type": ["visible", "hidden"], "description": "Whether the bibliography is 'shown' or 'hidden'. \
```

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+        "cite" => transform_cite(to, &input),
+        "cite-internal-do-not-use" => transform_cite_label(to, &input),
+        "bibliography" => transform_bibliography(to, &input),
+        other => {
+            eprintln!("Package does not support {other}");
+        }
+    }
+}
+
+// The name Citation clashes with Hayagriva
```
adelhult, line 135: I think we can remove the comment since this name is fine too : ). Or perhaps import `Citation as HayagrivaCitation`
```suggestion
```

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+fn transform_bibliography(_to: &str, input: &Value) {
+    let Some(bibliography) = read_bibliography(input) else { return; };
+
+    // This is the citations (i.e [cite]s) used in the text
+    let citations: Vec<String> = {
+        let var = env::var("inline_citations").unwrap_or("[]".to_string());
+        serde_json::from_str(&var).unwrap()
+    };
+
+    // This is our database of all entries (like a db containing all biblatex-entries)
```
adelhult, line 182: This can contain yaml entries too right?
```suggestion
    // This is our database of all entries
```

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+    for citation_str in &citations {
+        // We parse the citation (which is a JSON obj)
+        let citation: InlineCitation = serde_json::from_str(citation_str).unwrap();
+
+        // We find the record by the key
+        let entry = if let Some(record) = database.records.get(citation.key) {
+            record.entry
+        } else {
+            eprintln!(
+                "Missing citation {0}, mentioned in [cite] {0} but not defined in the bibliography",
```
adelhult, line 202: Added `'` around the key and remove `[cite] ...` since we don't know how they have formatted it in the source code
```suggestion
                "Missing citation key '{0}', consider adding it to your  bibliography",
```

packages/bibliography/tests/test_bibliography.json
```diff
@@ -0,0 +1,70 @@
+{
+    "name": "bibliography",
```
adelhult, line 2: Nice with all the tests :)

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+    };
+}
+
+macro_rules! text {
+    ($e:expr) => {
+        module!("__text", $e)
+    };
+}
+
+fn manifest() {
```
adelhult, line 49: Looks good!

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+            citation_str,
+            format!("{json_citation}")
+        ));
+    }
+
+    let visibility = input["arguments"]["visibility"].as_str().unwrap();
+
+    // If the bibliography should be shown, show it!
+    if visibility == "shown" {
+        for entry in database.bibliography(bibliography_style.as_ref(), None) {
```
adelhult, line 250: I think the bibliography looks good but I think we should consider if we can style it some more (while keeping it flexible regarding the output format of course). For instance, the bracket with the numbers should end up in a column of the own instead of inside the text.

Also, it looks like it breaks on LaTeX output:
(But this looks to be my fault with the missing  `\n`  in the [newline] module that leads to TeX parsing the brackets as an argument to `\\` :sweat_smile: )
image

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+        }
+    }
+
+    println!("{}", Value::Array(output));
+}
+
+/// Gets the styles for the given key, as a pair of the bibliography style and citation style. If
+/// the key has no style mappings, None is returned. For the keys IEEE, APA, MLA and CMoS, Some is
+/// always returned, and for any other key, None is returned.
+fn get_styles(key: &str) -> Option<(Box<dyn BibliographyStyle>, Box<dyn CitationStyle>)> {
```
adelhult, line 268: Looks good!

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+/// * Trying to parse it as a BibLaTeX file
+/// * If that doesn't work, try to parse it as a Hayagriva Yaml file
+///
+/// If everything succeeds, the list of entries is returned, otherwise errors are
+/// printed to stderr and None is returned. Parsing errors will print both BibLaTeX
+/// and Yaml errors.
+fn read_bibliography(input: &Value) -> Option<Vec<Entry>> {
+    let filename = input["arguments"]["file"].as_str().unwrap();
+    let input_string: Cow<'_, str> = if filename.is_empty() {
+        Cow::Borrowed(input["data"].as_str().unwrap())
```
adelhult, line 302: Nice that you took the care to use Cow's!

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+        "CMoS" => Box::new(ChicagoAuthorDate::new()),
+        _ => return None,
+    };
+
+    Some((bibliography_style, citation_style))
+}
+
+/// This function reads the bibliography by first:
+/// * Checking the 'file' argument, and if it is non-empty, reads that file
+/// * Otherwise, checking the body of the module
```
adelhult, line 290: Suggestion: perhaps we could append the content of the [bibliography] to the file's content if both are defined instead of just ignoring one of them.

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+            None
+        }
+    }
+}
+
+/// This function displays a `DisplayString` by converting its formatting rules/spans
+/// to JSON representations that can be picked up by `ModMark` again. It uses
+/// __text for all text, wrapped in __bold and __italic for those rules, and uses
+/// the `[link]` package for URLs
+fn display(string: &DisplayString) -> Vec<Value> {
```
adelhult, line 333: Sounds good!

packages/bibliography/src/main.rs
```diff
@@ -0,0 +1,358 @@
...
+        let citation = InlineCitation { key, note };
+        serde_json::to_string(&citation).unwrap()
+    };
+
+    let output = json!([push_citation!(&citation), cite_label!(&citation)]);
+
+    println!("{output}");
+}
+
+fn transform_cite_label(_to: &str, input: &Value) {
```
adelhult, line 158: I think it would be really nice to use the `[reference]` and `[label]` modules so that you can click on in-text citation:s too. But, we can of course leave that for later : )

CMDJojo: I thought of this, however I am unsure if [label] (for LaTeX) works anywhere. I thought it needed to be just after a figure or something. I can experiment a bit with this! But yeah, it would be very nice to have that indeed.

packages/bibliography/tests/test_bibliography.json
CMDJojo: Thanks! The tests are far from flawless since we don't set env on tests (maybe we should add support for that to the test runner). This leads to this test not including `[1]` etc, even though it should, since the sources are never cited (this is how Hayagriva returns uncited sources; without counters)

packages/bibliography/src/main.rs
CMDJojo: Yes, it may indeed contain yaml entries. This was more of an explanation, "this is our database, behaving like a database of all biblatex-entries" to understand what it was a database of and get a better grasp on what it is doing. But I'll remove that part of the comment, sure!

packages/bibliography/src/main.rs
CMDJojo: I don't know why it doesn't render. I dont think it is an issue with `\\` since when adding a newline it still isn't shown. It works in math mode `$[2]$` but that seems unintended.

adelhult: Yeah, but the problem is that newlines are allowed to consume a pair of brackets as an option `\\[3cm]`

CMDJojo: Oh, didn't know about the newline option. Can we change newline to be `\\[]`?

adelhult: Or just `\\\n`

CMDJojo: It didn't work for me when adding a newline after `\\`. Does it for you? Otherwise, `\\[0pt]` works hehe

CMDJojo: `\\{}` works, that may be prefferable

packages/bibliography/src/main.rs
CMDJojo: Hmm, I don't know if this would be good and useful or just confusing to users. But that is a nice suggestion, maybe we should think about that some more.

CMDJojo: Added a warning if we supply both a body and file. We can change this behaviour later on if needed.

adelhult (approved):

PR #257, opened 2023-04-25
Replace playground with a React project

### LINKED ISSUE #256: Improve playground (react rewrite)
We should improve the playground and add more pages to the website with a home page as well as documentation. It might be best to rewrite this using React.

adelhult:
Resolves #256. A full rewrite of the playground in React. Feel free to play around with it :). There are still a bunch missing but I think we might as well rebase into main so that I'm not blocking others from contributing.

I also made a few issues for future improvements that we can work on if we have time or think that it would be fun:
* #302
* #303 
* #304 
* #305


.github/workflows/DeployPreview.yml
```diff
@@ -27,35 +27,36 @@ jobs:
...
-        run: chmod +x ./build-playground.sh
-      - name: Build Playground
-        run: ./build-playground.sh
+      - name: Build website
+        run: |
+          npm ci
+          chmod +x ./build_website.sh
+          export BASE="/pr-preview/pr-${PR}" && ./build_website.sh
+        env:
+          PR: ${{ github.event.number }}
```
CMDJojo, line 46: You can just set `BASE` directly:
```suggestion
          BASE: /pr-preview/pr-${{ github.event.number }}
```
Would look cleaner

.github/workflows/DeployMain.yml
```diff
@@ -1,5 +1,5 @@
 # This is a workflow based on an example provided by GitHub and on rossjrw/pr-preview-action
-name: Deploy main Playground
+name: Deploy main website
```
CMDJojo, line 2: It is fine to rename it, but please do rename the name of the preview workflow as well!

README.md
```diff
@@ -24,7 +24,7 @@ add new //modules// to the language, which adds functionality to the
 language, like [link label="links"](https://modmark.org), lists and more!
 ```
 
-The `[link]` syntax invokes a **module**, which lives outside the language itself, and the ability to use custom modules enhances the language and gives it capabilities way beyond what one single syntax set may. Make sure to check out the [playground](https://modmark.org) to see an example document which uses modules!
+The `[link]` syntax invokes a **module**, which lives outside the language itself, and the ability to use custom modules enhances the language and gives it capabilities way beyond what one single syntax set may. Make sure to check out the [playground](https://modmark.org/#&playground) to see an example document which uses modules!
```
CMDJojo, line 27: Is this the correct URL? It seems like a strange-looking URL to me, I would think it would be `https://modmark.org/playground`. But, I don't know much about react routing

website/.vite/deps_temp_39fa3b02/package.json
```diff
@@ -0,0 +1,3 @@
+{
+  "type": "module"
+}
```
CMDJojo, line 3: Should this file be commited? I don't know much about vite, but it is under `.vite` AND in a directory called temp

website/src/Homepage.tsx
```diff
@@ -0,0 +1,309 @@
...
+        color="#816796"
+      >
+        <h1>
+          <span style={{ opacity: 0.4 }}>#</span> Familiar and <span style={{ opacity: 0.4 }}>**</span>powerful<span style={{ opacity: 0.4 }}>**</span>
+        </h1>
+        <p>
+          If you have used Markdown before you will quickly feel right at home.
+        </p>
+        <p>
+          Read the getting started guide to learn more about different ways to <CodeFont>**style** ==text==</CodeFont> and include other elements in your document like a <CodeFont>[image](./image.png)</CodeFont> or a <CodeFont>[link](https://modmark.org)</CodeFont>.
```
CMDJojo, line 256: It looks super cool, with all these codefonts and such, and I like that it looks kind of similar to markdown as an introduction. But, maybe, we shouldn't use `[image]` as an example since it doesn't really work in-line...

website/src/main.tsx
```diff
@@ -0,0 +1,80 @@
...
+function PrPrompt() {
+  const [hidden, setHidden] = useState(false);
+
+  const location = window.location;
+  const regex = /pr-preview\/pr-(\d+)/;
+  console.log(location.href);
+  const match = location.href.match(regex);
+  if (match !== null) {
+    return hidden ? <></> : <DebugMessage>
+      <a href={`https://github.com/modmark-org/modmark/pull/${match[1]}`}>Preview of PR #{match[1]}</a>
```
CMDJojo, line 41: I love this message! But for some reason, I think the styling is wrong; when you go to the playground without this header you can't scroll (like, you can scroll in the text fields but not scroll down the entire page, which is reasonable), but this header makes it possible to "scroll away" the header. I don't know the reason for this, but maybe the page is as high as the monitor, regardless of if there is a bar or not?

website/src/welcomeMessage.tsx
```diff
@@ -0,0 +1,4 @@
+const welcomeMessage = `# ModMark
+Hello and welcome to the ModMark playground!
+`;
+export default welcomeMessage;
```
CMDJojo, line 4: Can't we use the previous document as a welcome message? It includes some more stuff to play around with, and was easy for us to handle since it was in a separate file and such

CMDJojo (commented):
Nice work with the rewrite! I know you are not quite done yet, but I left some comments on stuff you might have missed! Keep it up :) it looks so good!

CMDJojo:
Also, I noticed that it says "Typing..." as soon as you load the page (when I know that can't be right). Might be configured so that that is displayed when the WASM binary is loaded, but it would be nice if it said like "Loading compiler..." or something that isn't a lie haha

hugomardbrink (approved):
LGTM!

PR #261, opened 2023-04-25
GH-264: Add `[hide]` and `[comment]` modules

CMDJojo:
Resolves GH-264 by adding `[hide]` and `[comment]` modules, both of which just doesn't do anything with its input


CMDJojo:
Closed due to wrong branch name

PR #276, opened 2023-04-27
Add latex support to the code module

bruhng:
Resolves: GH-113
Adds support to the code module for latex

### DIFF
--- a/packages/code/src/main.rs
+++ b/packages/code/src/main.rs
@@ -4,11 +4,10 @@ use std::io::{self, Read};
 
 use serde_json::{json, Value};
 use syntect::easy::HighlightLines;
-use syntect::highlighting::{Color, ThemeSet, Theme, Style};
+use syntect::highlighting::{Color, ThemeSet, Theme};
 use syntect::html::{styled_line_to_highlighted_html, IncludeBackground};
 use syntect::parsing::{SyntaxSet, SyntaxReference};
 
-use syntect::util::{as_latex_escaped,LinesWithEndings};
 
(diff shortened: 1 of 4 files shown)


packages/code/src/main.rs
```diff
@@ -4,9 +4,15 @@ use std::io::{self, Read};
...
 use serde_json::{json, Value};
 use syntect::easy::HighlightLines;
-use syntect::highlighting::{Color, ThemeSet};
+use syntect::highlighting::{Color, ThemeSet, Theme};
 use syntect::html::{styled_line_to_highlighted_html, IncludeBackground};
-use syntect::parsing::SyntaxSet;
+use syntect::parsing::{SyntaxSet, SyntaxReference};
+
+
+const VERBATIM_FIX : &str = r"\makeatletter
```
CMDJojo, line 12: Maybe a more descriptive name?

packages/code/src/main.rs
```diff
@@ -94,13 +120,118 @@ fn transform_code(to: &str) {
...
+            } 
+        }
         other => {
             eprintln!("Cannot convert code to {other}");
         }
     }
 }
 
-fn get_style(
+fn replace_tabs(s: &str, tab_size: u64) -> String {
```
CMDJojo, line 141: I think we can just do `s.replace('\t', " ".repeat(tab_size))`

CMDJojo (approved):
Looks good to me! Nice to get LaTeX support :)

Left some small notes, but overall I think the code is good

Issue #275, opened 2023-04-27
Math tags sometimes crashes (in HTML)


CMDJojo:
`$$x^2$$` fails to compile sometimes

PR #277, opened 2023-04-27
GH-275: Add `evaluate-before-children` option to transforms

CMDJojo:
This PR resolves GH-275. The problem was that `$$x^2$$` gets represented as `__math { __text "x^2" }`, and if `__text` is evaluated before `__math`, it won't work since `__text` will be replaced with just a raw string. The core issue is that we have no dependency saying that `__math` must be evaluated before `__text`, as is the case here to get correct output (and it is also the case with `__verbatim`). This PR adds an option to the transforms in the manifest `evaluate-before-children`, default to `false`, that when true, if a parent is detected with the corresponding name, it will be guaranteed to be evaluated before its children. This PR also sets this flag for `__math` for HTML/LaTeX and `__verbatim` for LaTeX (unimplemented in HTML). This flag should only be used when the parent captures its children and changes them (doesn't necessarily include its children in its own output).

This PR is ready for review immediately.

### DIFF
--- a/core/src/context.rs
+++ b/core/src/context.rs
@@ -209,6 +209,7 @@ where
 pub struct Dependencies {
     pub var_accesses: Vec<((String, VarType), VarAccess)>,
     pub has_unknown_content: bool,
+    pub evaluate_before_children: bool,
 }
 
 impl<T, U> Context<T, U> {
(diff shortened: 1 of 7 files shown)


bruhng (approved):
Great fix, weird bug lol

Issue #280, opened 2023-05-02
Resolving local packages works differently in playground/CLI


axoxelol1:
Local imports like
```
[config]
import main
```
work differently in the cli and playground. The cli will look for `main.wasm` while playground looks for `main`. We should either make the playground also add the `.wasm` extension or the cli should not add it. Not sure what is best but they should be the same at least.

CMDJojo:
Great lifting this issue! One technical note is that the name including extension is wrapped in the enum the Resolver gets to resolve.

I would say that the best way to solve this is:
* Choose an extension for our packages (we use `wasm` know, we may stick with it or move to `mmpkg` or whatever we like)
* If the string to resolve ends in `.<extension>`, keep it as-is, otherwise append it

This will behave similar to many applications which, if you want to store a file as pdf, it appends `.pdf` if absent, even if the file name already has a file extension.

We might want to add the logic for this to Core so it has more control of how stuff are being loaded, so the behaviour is easier synced. Also, worth taking into account, it will with this change be impossible to import a package with the wrong extension. I think it is fine and a reasonable thing to assert, but opinions may differ.

PR #283, opened 2023-05-02
GH-280: Append `.wasm` to local packages in modmark-core

CMDJojo:
This PR resolves GH-280 by making Core check if the file name ends in `.wasm` and, if it doesn't, append it to the file name. This means that regardless if the user specifies `import foo` or `import foo.wasm`, Core will give a `ResolveTask` saying "*Resolve the file `foo.wasm`*" and the resolvers themselves doesn't have to add the extension. This means that more logic has been moved from the split-up resolvers to the same Core they both use, making unsynchronized behaviour more unlikely.

### DIFF
--- a/cli/src/package.rs
+++ b/cli/src/package.rs
@@ -140,16 +140,12 @@ impl PackageManager {
     }
 
     fn fetch_local(&self, package_path: &str) -> Result<Vec<u8>, CliError> {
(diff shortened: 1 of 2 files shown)


axoxelol1 (approved):
Looks good and works as expected. Nice to have that functionality shared instead so that they won't behave differently. Nice work :)

PR #284, opened 2023-05-03
GH-198: Ignore WASI errors during import if exit code was 0

axoxelol1:
This is connected to issue #198 only for imports/reading manifest. When creating a ZIG package I kept getting the error when importing a package despite not doing anything other than printing a manifest to stdout. An equivalent package in Rust had no problems, I also checked to see that they printed identical json.

With the early exits as in issue #198 I read that it could be because the call stack not being emptied but I have not had a lot of luck finding information on the issue. When googling again for this issue I found https://github.com/wasmerio/wasmer/issues/2566 which seems to describe the problem we have. It seems like the issue is with the Rust create API and that the way it uses error is more like a message.

### DIFF
--- a/core/src/package.rs
+++ b/core/src/package.rs
@@ -448,8 +448,16 @@ impl Package {
         wasi_env.data_mut(store).set_memory(memory.clone());
 
         // Retrieve manifest of package
-        let manifest = instance.exports.get_function("_start")?;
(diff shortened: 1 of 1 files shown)


Forsinge (approved):
Nice job, seems to work! Looks similar to the check in `transform_from_wasm` which is already merged.

Issue #285, opened 2023-05-03
Do not use string comparison for checking exit code 0


CMDJojo:
This code is used both in `context.rs` when doing a WASM transformation and in `package.rs` when reading the manifest:

```rs
// TODO: See if this can be done without string comparison
let error_msg = e.to_string();
if !error_msg.contains("WASI exited with code: 0") {
  ...
}
```

It would be great to not have a string comparison here, since future versions of wasmer may change the stringification which will make this comparison break. It is better to actually check the error code.

PR #286, opened 2023-05-03
GH-285: Check WasiError::Exit(0) instead of string comparison

CMDJojo:
This changes the string comparison for checking exit code 0 to an attempt at downcasting to WasiError::Exit. Resolves GH-285


axoxelol1 (approved):
Sweet, this works as expected and is hopefully more future proof.

Issue #287, opened 2023-05-03
Code skeletons for developing packages


hugomardbrink:
**Is your feature request related to a problem? Please describe.**
It is hard to know where to start when developing packages, templates for different languages should help this process.

**Describe the solution you'd like**
A subcommand that initialises a package code skeleton for different languages

PR #288, opened 2023-05-03
GH-287: Add subcommand for initialising package code skeletons

hugomardbrink:
- Resolves: GH-287.

### DIFF
--- a/cli/src/error.rs
+++ b/cli/src/error.rs
@@ -47,4 +47,7 @@ pub enum CliError {
 
     #[error("Second argument OUTPUT_FILE missing. You may only omit this when compiling to html and using the live preview.")]
     MissingOutputFile,
+
+    #[error("Could not resolve template tag: '{0}'.")]
+    TemplateTag(String),
 }
--- a/cli/src/main.rs
+++ b/cli/src/main.rs
@@ -1,7 +1,7 @@
 use std::{
     collections::HashMap,
     env,
-    fs::{self, File},
+    fs::{self, remove_dir_all, remove_file, File},
     io::{stdout, Write},
     path::{Path, PathBuf},
     sync::{
@@ -149,6 +149,15 @@ impl CompileArgs {
     }
 }
 
+#[derive(Parser)]
+struct InitArgs {
+    #[arg(
+        index = 1,
+        help = "What language the template will be made in. Enter 'list' for a list of languages."
+    )]
+    language: String,
+}
+
 #[derive(Subcommand)]
 enum Command {
     Compile(CompileArgs),
@@ -156,6 +165,7 @@ enum Command {
         #[command(subcommand)]
         command: CacheCommand,
     },
+    Init(InitArgs),
 }
 
 #[derive(Subcommand)]
@@ -340,6 +350,15 @@ async fn main() {
                     .unwrap();
             }
         },
+        Command::Init(init_args) => match run_init(init_args).await {
+            Ok(_) => (),
+            Err(error) => {
+                let mut stdout = stdout();
+                stdout
+                    .execute(style::PrintStyledContent(format!("{error}").red()))
+                    .unwrap();
+            }
+        },
     }
 }
 
@@ -461,6 +480,54 @@ async fn run_cache(command: &CacheCommand) -> Result<(), CliError> {
     }
 
     Ok(())
+}
+
+async fn run_init(args: &InitArgs) -> Result<(), CliError> {
+    let template_json = include_str!("templates.json");
+    let mappings: HashMap<String, Vec<String>> = serde_json::from_str(template_json)?;
+
+    match args.language.to_lowercase().trim() {
+        "list" => {
+            println!("These are the available languages:");
+            mappings
+                .iter()
+                .for_each(|(_, languages)| println!("\t{}", languages[0]));
+            Ok(())
+        }
+        language => {
+            let language = language.to_string();
+
+            let Some(url) = mappings
+        .iter()
+        .find_map(|(k, v)| v.contains(&language).then_some(k)) else {
+            return Err(CliError::TemplateTag(language));
+        };
+
+            let Some(folder_name) = url.rsplit_once('/').map(|(_, second)| second) else {
+            return Err(CliError::TemplateTag(language));
+        };
+
+            std::process::Command::new("git")
+                .arg("clone")
+                .arg(url)
+                .output()?;
+
+            let project_path = fs::canonicalize(folder_name)?;
+
+            let mut git_path = project_path.clone();
+            git_path.push(".git");
+
+            let mut gitignore_path = project_path.clone();
+            gitignore_path.push(".gitignore");
+
+            remove_dir_all(&git_path)?;
+            remove_file(&gitignore_path)?;
+
+            println!("{}{}", "Package template created in: ", folder_name);
+
+            Ok(())
+        }
+    }
 }
 
 /// Choose a free port for hosting the html live preview
--- a/cli/src/templates.json
+++ b/cli/src/templates.json
@@ -1,0 +1,10 @@
+{
+  "https://github.com/modmark-org/rust_template": ["rust", "rs"],
+  "https://github.com/modmark-org/cpp_template": ["c++", "cpp"],
+  "https://github.com/modmark-org/c_template": ["c"],
+  "https://github.com/modmark-org/assemblyscript_template": [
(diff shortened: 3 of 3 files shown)


cli/src/main.rs
```diff
@@ -463,6 +483,43 @@ async fn run_cache(command: &CacheCommand) -> Result<(), CliError> {
     Ok(())
 }
 
+async fn run_init(args: &InitArgs) -> Result<(), CliError> {
+    match args.language.to_lowercase().trim() {
+        "rust" => {
```
CMDJojo, line 488: Maybe alias as rs also?

cli/src/main.rs
```diff
@@ -463,6 +483,43 @@ async fn run_cache(command: &CacheCommand) -> Result<(), CliError> {
...
+        "c" => {
+            println!(
+                "{:?}",
+                std::process::Command::new("git")
+                    .arg("clone")
+                    .arg(C_TEMPLATE)
+                    .output()
+            );
+        }
+        _ => return Err(CliError::Cache),
```
CMDJojo, line 515: Maybe not `CliError::Cache`? We should probably do a custom error type, like `CliError::TemplateTag("swift")` to get the error `Template matching "swift" doesn't exist`

cli/src/main.rs
```diff
@@ -463,6 +483,43 @@ async fn run_cache(command: &CacheCommand) -> Result<(), CliError> {
     Ok(())
 }
 
+async fn run_init(args: &InitArgs) -> Result<(), CliError> {
+    match args.language.to_lowercase().trim() {
```
CMDJojo, line 487: I think the way to go here may not be to actually have a `match` here that includes hard-coded URLs, it would probably be best to have a map of sorts, like we do for `package-registry`. I am fully aware that for three projects this works fine, and it seems unnecessary to complicate things further in this early stage, but can we go half-way and like `include_str!` a file containing JSON mappings to things? Like if we have a file `templates.json`:
```json
{
    "https://github.com/modmark-org/rust_template": ["rust", "rs"],
    "https://github.com/modmark-org/cpp_template": ["c++", "cpp"],
    "https://github.com/modmark-org/c_template": ["c"]
}
```
, then `let mappings: HashMap<String, Vec<String>> = serde_json::from_json(include_str!("templates.json"));` statically once, and do searches like `mappings.iter().find(|(k, v)| v.contains(args.language.to_lowercase().trim()).then_some(k))`

CMDJojo (commented):
Good job with the implementation! I think package templates are a very important thing to help developers out starting to develop their package

cli/src/main.rs
```diff
@@ -463,6 +483,43 @@ async fn run_cache(command: &CacheCommand) -> Result<(), CliError> {
...
+                    .output()
+            );
+        }
+        "c++" | "cpp" => {
+            println!(
+                "{:?}",
+                std::process::Command::new("git")
+                    .arg("clone")
+                    .arg(CPP_TEMPLATE)
+                    .output()
```
CMDJojo, line 503: (This comment didn't appear in my review for some reason)

`git clone`ing is smart here, but I think we should remove the `.git` directory since we don't really want our template being a git repo. We can do this in two ways:
* Keep it `git clone`, then `cd` into the dir and `rm -r .git`/`rmdir .git`
* Download it as a zip file and unzip it

cli/src/main.rs
hugomardbrink: Oops, youre right, this was a placeholder heh

axoxelol1 (changes requested):
Nice work, this is very useful. I would like to see a list of available formats when using `modmark cache --help` or something similar. Right now the only way to find the options seems to be to look in the json file.

CMDJojo:
I am unsure if `rm -rf` works on Windows, I think we need to use `rmdir` instead. However, I am very unsure. Maybe @Forsinge can try it out on Windows?

Forsinge:
It seems like @CMDJojo is correct. `rm -rf` does not work for me (Windows 10). I think the full command to use on Windows is `rd /q /s` or `rmdir /q /s` (both of which work for me).

hugomardbrink:
> > It seems like @CMDJojo is correct. `rm -rf` does not work for me (Windows 10). I think the full command to use on Windows is `rd /q /s` or `rmdir /q /s` (both of which work for me).
> 
> Yeah, although this would work I think I will probably use something like this instead: https://doc.rust-lang.org/std/fs/fn.remove_dir_all.html

Seems more robust :)

cli/src/main.rs
```diff
@@ -149,13 +149,23 @@ impl CompileArgs {
     }
 }
 
+#[derive(Parser)]
+struct InitArgs {
+    #[arg(
+        index = 1,
+        help = "What language the template will be made in. Currently available: c++, c and rust."
```
CMDJojo, line 156: You can probably add "list" as a special *language* which just lists all available languages. Regardless if you do it this way or with a flag, I think hard-coding is bad since then we need to keep it in sync with templates.json, and if we move that to a separate repo, we would need to change ModMark to facilitate adding new templates.

```suggestion
        help = "What language the template will be made in. Enter 'list' for a list of languages."
```

cli/src/main.rs
```diff
@@ -463,6 +482,53 @@ async fn run_cache(command: &CacheCommand) -> Result<(), CliError> {
...
 }
 
+async fn run_init(args: &InitArgs) -> Result<(), CliError> {
+    let template_json = include_str!("templates.json");
+    let mappings: HashMap<String, Vec<String>> = serde_json::from_str(template_json)?;
+    let language = args.language.to_lowercase().trim().to_owned();
+
+    let Some(url) = mappings
+        .iter()
+        .find_map(|(k, v)| v.contains(&language).then_some(k)) else {
```
CMDJojo, line 492: You probably won't need the string owned
```suggestion
    let language = args.language.to_lowercase().trim();

    let Some(url) = mappings
        .iter()
        .find_map(|(k, v)| v.contains(language).then_some(k)) else {
```

cli/src/main.rs
```diff
@@ -463,6 +482,53 @@ async fn run_cache(command: &CacheCommand) -> Result<(), CliError> {
...
+    std::process::Command::new("rm")
+        .current_dir(&project_path)
+        .arg("-rf")
+        .arg(".git")
+        .output()?;
+
+    std::process::Command::new("rm")
+        .current_dir(&project_path)
+        .arg(".gitignore")
+        .output()?;
```
CMDJojo, line 525: These aren't needed since you added remove_file and remove_dir_all, right?

```suggestion
```

cli/src/main.rs
```diff
@@ -463,6 +482,53 @@ async fn run_cache(command: &CacheCommand) -> Result<(), CliError> {
...
+
+    println!("Package template successfully intitialised!");
```
CMDJojo, line 527: Maybe point to the path, like `Package template created in the folder cpp_template`? Since you already have the folder name available

CMDJojo (commented):
Good work! It's looking better now :) I think this is ready to merge soon

axoxelol1 (approved):

axoxelol1:
Whoops did not mean to approve, don't merge until @CMDJojo's comments has been fixed heh

cli/src/main.rs
```diff
@@ -463,6 +482,54 @@ async fn run_cache(command: &CacheCommand) -> Result<(), CliError> {
...
+            Ok(())
+        }
+        _ => {
+            let language = args.language.to_lowercase().trim().to_string();
+
+            let Some(url) = mappings
+        .iter()
+        .find_map(|(k, v)| v.contains(&language).then_some(k)) else {
+            return Err(CliError::TemplateTag(language));
+        };
```
CMDJojo, line 504:
```suggestion
        language => {
            let Some(url) = mappings
        .iter()
        .find_map(|(k, v)| v.contains(language).then_some(k)) else {
            return Err(CliError::TemplateTag(language));
        };
```

cli/src/error.rs
```diff
@@ -47,4 +47,7 @@ pub enum CliError {
 
     #[error("Second argument OUTPUT_FILE missing. You may only omit this when compiling to html and using the live preview.")]
     MissingOutputFile,
+
+    #[error("Could not resolve template tag: '{0}'.")]
```
CMDJojo, line 51: Maybe "No template named '{0}'" for a little more user-friendly look, but not a blocker

Issue #292, opened 2023-05-04
Future incompatibility due to Warp


CMDJojo:
When compiling (or checking) the CLI, we get this warning from Cargo:
> warning: the following packages contain code that will be rejected by a future version of Rust: buf_redux v0.8.4, multipart v0.18.0

Both of these dependencies are brought in by `warp 0.3.3` and it can be solved by bumping Warp to 0.3.5

PR #293, opened 2023-05-04
GH-292: Bump warp to 0.3.5

CMDJojo:
Resolves GH-292


axoxelol1 (approved):
Nice catch, update does not seem to break anything =)

CMDJojo:
"Update doesn't seem to break anything" is probably the best comment I've ever gotten!

Issue #294, opened 2023-05-08
Importing packages from GitHub releases does not work in playground


Forsinge:
**Describe the bug**
Many of the packages in the package catalog (https://github.com/modmark-org/package-registry) provide .wasm files as part of a release. Because of CORS restrictions the playground is unable to fetch these packages.

**To Reproduce**
1. Open playground
2. Write a config module and import catalog:prettify

PR #295, opened 2023-05-08
Use proxy to fetch packages from catalog in playground

Forsinge:
Resolves: #294 

This PR solves the CORS problem by using a CloudFlare worker as a proxy, which adds the necessary `Access-Control-Allow-Origin` header. I used the code at https://developers.cloudflare.com/workers/examples/cors-header-proxy with the addition of manually sending another request to a new URL if the response status was 302.


CMDJojo (approved):
Super nice! No need for cors-bypasses in the browser anymore 🎉

Issue #296, opened 2023-05-09
Use inline-content for list content


CMDJojo:
We should use inline-content in lists rather than raw, so that we can style our content and so that text will be escaped.

```
[list]
1. **Item 1**

[comment]
The example above should have bold text "Item 1"

[list]
1. <script>alert(1)</script>

[comment]
This example should not do an alert
```

PR #297, opened 2023-05-09
GH-296: Use inline-content for list items

CMDJojo:
Resolves GH-296, and also does some small other general improvements

### DIFF
--- a/packages/list/src/lib.rs
+++ b/packages/list/src/lib.rs
@@ -10,12 +10,6 @@ macro_rules! import {
     ($package:expr) => {module!("set-add", $package, {"name": "imports"})}
 }
 
-macro_rules! raw {
-    ($package:expr) => {
-        module!("raw", $package, {})
-    };
-}
-
 macro_rules! inline_content {
     ($package:expr) => {
         module!("inline_content", $package, {})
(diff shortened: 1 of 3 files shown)


packages/list/src/lib.rs
```diff
@@ -2,8 +2,24 @@ use std::str::FromStr;
...
+macro_rules! module {
+    ($name:expr, $data:expr $(,$($args:tt)*)?) => {json!({"name": $name $(,"arguments":$($args)*)*, "data": $data})}
+}
+
 macro_rules! import {
-    ($e:expr) => {json!({"name": "set-add", "arguments": {"name": "imports"}, "data": $e})}
+    ($package:expr) => {module!("set-add", $package, {"name": "imports"})}
+}
+
+macro_rules! raw {
```
adelhult, line 13: Can't we just use json string literals instead of the raw macro

adelhult (approved):

packages/list/src/lib.rs
CMDJojo: Yes, we can! I just refactored the code as it was without changing it, but I will change to the raw macro before merging

Issue #289, opened 2023-05-04
Images with width less than 1.0 can not be centered


Forsinge:
**Describe the bug**
A likely explanation is that the surrounding figure becomes centered but the width of the image inside is only scaled to the left. If this is the case, it is an issue with [plot] and similar modules as well.

**Screenshots**
![image](https://user-images.githubusercontent.com/89988910/236146423-db83f480-b340-461a-bd71-3661c902cbf3.png)

PR #306, opened 2023-05-10
Fix centering for images

Forsinge:
Resolves: #289

This PR also removes usage of the `raw!` macro.


CMDJojo:
image
it seems like the "figure" is as large as it was before, and this change centering the image will **always** center it in the page. I think the thing to do is to shrink the figure rather than just the image

bruhng (approved):
Tried it and it works great! Good job!

Issue #307, opened 2023-05-10
Add option to restrict transform to multiline-invokations only


CMDJojo:
Currently, all parent and module transforms gets propagated to the package responsible for executing the transform. When we implemented custom data types, we moved some logic and verification from the package itself to core, yielding better error messages and resulting in Core not having to bother the package since it already knew before-hand that something was wrong. I think this change was very good.

Now I propose a new addition to the checks Core will do. For transformations in general, as long as the arguments match, the node is sent to the package. We could add a `type` field that declare what node types the transform accepts, like if it accepts a `parent`, a `module` or both. This will ensure that `[__document] foo` isn't sent to the HTML package, which messes up rendering, and will in many cases avoid crashes where a package only expects parent transforms and force-unwraps those fields. This will thus differentiate more between `parent` and `module` transforms, and would resolve #208.

In addition to this, modules may be invoked inline or multiline. If it is an inline invocation, it should by convention only generate inline content, and if it is an multiline invocation, it should by convention only generate block content. This works fine for the most part, and most packages respects this (such as `[code]` being displayed very differently in the different cases). However, there are some situations where only one of the kinds are reasonable, like for `[image]` and `[table]`. It doesn't really make sense to have an inline image, nor does it make sense to have an inline table.

My proposal to a solution is to add a `type` field, which can be one of five enums:
* `module` (default) - the transformation accepts both inline and multiline modules
* `inline-module` - the transformation only accepts inline modules
* `multiline-module` - the transformation only accepts multiline modules
* `parent` - the transformation only accepts parents
* `any`/`*` - the transformation accepts both inline and multiline modules, as well as parents

The reason for `module` being the default is that it is the most commonly-used type, and would ease development of packages. If someone is to develop their first package, it would probably be for a module transformation, and it is nice to be able to do that without very much knowledge about the system. If someone is developing a language package, they probably have more knowledge about the system, and thus the type field won't be much of a hassle to use.

I don't know if `any` is of particular interest, but in theory someone could make `[__strikethrough]` invokeable both as a module and still support it as a parent by checking if it has a `children` or a `data` field.

I think this change will make for better error handling and more self-documenting packages. This enum will of course be included in the data sent to the website so it can use it when displaying a list of transforms.

Issue #208, opened 2023-04-04
Make module and parent transforms more distinct


CMDJojo:
**Is your feature request related to a problem? Please describe.**
There are currently two types of transforms, module transforms that transforms module expressions like `[math] x^2` and parent transforms that transforms a parent node, which is an internal representation containing a list of other nodes (children). This isn't a problem per se, but the language itself doesn't create much of a distinction between them; transforms are only identified by name and not by type, and the only actual difference between them is how the JSON sent to the module looks like. That itself is more of a coincidence, that `__bold` transforms *usually* refer to parent transforms and `math` transforms *usually* refer to module expressions, and the JSON serialization just happens to be in a different structure between those two, and there is nothing interent in the language itself that it is guaranteed to work (nor is it). Since transforms are only identified by name, one could enter `[__document]` into the document and see everything go bad. Modules that expect one thing to be a module and gets it as a parent may crash due to assertions it believed to be accurate, due to the lack of distinction. Additionally, it would make it more self-documenting, it would be easy to see that `__math` is a parent transform and that `math` is a module expression. Currently, developers may have a hard time even figuring out how to "call" the very important `raw` transform; is it a module or parent transform?

**Describe the solution you'd like**
I think we somewhere should make a larger distinction between what names refer to module expressions and what names refer to parent transforms (maybe introduce a new data type that also encode the transform type). That would make it so that if anyone enters `__document`, we wouldn't try to use the parent transform and just report that there is no module expression transform with that name. Additionally, it may allow us to overload names (so that there is a `[table]` module expression and a `table` parent transform), which we may decide on separately if that should be allowed or not.

**Additional context**
I would love to have a discussion about possible ways to implement this, here in the comments, with pros and cons. I will (probably) also follow up with a more concrete implementation example.

PR #308, opened 2023-05-10
Add ArgumentType

CMDJojo:
This PR resolves GH-307 and resolves GH-208 by adding an enum keyed `type` to `Transform`:
```rust
#[serde(rename_all = "kebab-case")]
pub enum TransformType {
    #[default]
    Module,
    InlineModule,
    MultilineModule,
    Parent,
    #[serde(alias = "*")]
    Any,
}
```
Then, before executing a transform, the element type is checked such that if Module is specified and the element is a Parent, it will be transformed to an error node instead. This is in line with argument type checks, whereas the compiler sometimes doesn't execute a transform if it is doomed to fail, but rather early returns a nicer error message.

It is unclear which of the current transforms should have what type, but since `Module` is default, all the language packages have been changed so many transforms (such as `__document`, `__strikethrough` etc) are of type `Parent`. Also, the thing that started the discussion which led to this change was that `[image]` can't reasonably produce in-line content, so `[image]` is changed to type `MultilineModule`.

You can test this out by going to the playground and testing this document:
```
[__document]
This should fail since __document should be a parent

Bla bla bla [image] foo.jpg this should fail since ``[image]`` produces a figure (block)

[image]
foo.jpg

[comment]
This should work tho
```

This PR is ready for review immediately

### DIFF
--- a/cli/src/file_access.rs
+++ b/cli/src/file_access.rs
@@ -35,12 +35,12 @@ impl AccessPolicy for CliAccessManager {
         !self.deny_create
     }
 
-    fn allowed_access(&mut self, path: &Path, module_name: &String) -> bool {
+    fn allowed_access(&mut self, path: &Path, module_name: &str) -> bool {
         if self.allow_every_module {
             return true;
         }
 
-        let module = self.modules.entry(module_name.clone()).or_default();
+        let module = self.modules.entry(module_name.to_string()).or_default();
         let path_str = path.to_str().expect("Could not convert path to &str");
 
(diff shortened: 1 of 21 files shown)


CMDJojo:
I also added type-like indicators to the package view to show what types the transform is for. Since 1. I'm not very good at web development and 2. The playground is undergoing a large rewrite anyways, I didn't spend very much time on it and it can surely be made to look much better.

image

Forsinge (approved):

Issue #313, opened 2023-05-11
Allow some parent-only elements to be invoked as a module


CMDJojo:
It would be nice if we can invoke, let's say, `__verbatim` as a module rather than just as a parent. This will allow for making larger verbatim blocks which spans over multiple lines, and contains line breaks.

PR #314, opened 2023-05-11
GH-313: Allow some parents to be invoked as modules

CMDJojo:
This PR resolves GH-313 by changing the type of the following transforms from `Parent` to `Any`:
* `__bold`
* `__italic`
* `__superscript`
* `__subscript`
* `__strikethrough`
* `__underlined`
* `__verbatim`

This applies to both the HTML and LaTeX package. Special care is taken for `__verbatim` so that it becomes `\verb||`/`<code>` when invoked inline, and `\begin{verbatim}`/`<pre>` when invoked multiline. This PR also resolves an issue where too many characters were escaped in `\verb`, so that ` ``hello_world`` ` was rendered `hello\_world`

This PR is ready for review immediately. This behaviour is also needed for writing the thesis.

### DIFF
--- a/packages/html/Cargo.toml
+++ b/packages/html/Cargo.toml
@@ -6,5 +6,5 @@ edition = "2021"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
-serde = {version = "1.0.152"}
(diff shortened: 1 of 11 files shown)


bruhng (approved):
Nice rewrite! fixes  the issues and makes it a lot easier to read

Issue #315, opened 2023-05-11
Add multi-line variant of table


CMDJojo:
It would be nice with a table module that supports mutli-line content, to facilitate larger tables like this one:
<img width="355" alt="image" src="https://github.com/modmark-org/modmark/assets/25929684/c92e8996-8668-4e52-b7af-e648c9720ff6">

PR #316, opened 2023-05-11
GH-315: Add big-table

CMDJojo:
This PR resolves GH-315 by adding a new `big-table` module. The PR also adds descriptions to the table modules. It also includes a test for the new module, and even though it would be nice with even more tests with higher coverage, I think the main focus should be the thesis right now and this module is needed there.

Here is an example document you can test out:
```
[big-table "Comparison between programming^TM languages - which is best?"]
## Python
[code py]
def foo():
    print("bar");
[next-column]
## Rust
[code rs]
fn foo() {
    println!("bar");
}
[next-row]
## Java
[code java]
public static void foo() {
    System.out.println("bar");
}
[next-column]
## ModMark
[code txt]
bar
```

### DIFF
--- a/core/src/lib.rs
+++ b/core/src/lib.rs
@@ -292,52 +292,94 @@ mod tests {
             version: "0.1".to_string(),
             description: "This package supports [table] modules".to_string(),
             transforms: vec![Transform {
-                from: "table".to_string(),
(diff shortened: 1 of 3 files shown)


axoxelol1 (approved):
Not sure about the default delimiters looking like modules, might be confusing. But we need it for the thesis and can change later if we feel a need to =)

Issue #323, opened 2023-05-12
Improve the styling of the bibliography module for LaTeX and HTML output


adelhult:
**Is your feature request related to a problem? Please describe.**
We should display the bibliography in a nicer way when we know how to (and of course offer a fallback to a simpler design for all other formats).

**Describe the solution you'd like**
The LaTeX ouput could look something like this:
![image](https://github.com/modmark-org/modmark/assets/11508459/70415355-0dab-46bf-962e-72863db5dd7a)

```latex
\renewcommand{\arraystretch}{1.5}
\section*{Bibliography}
\begin{tabularx}{\textwidth}{p{0.5cm} X}
[1] & \leavevmode \hypertarget{bibentry:djot}{J. MacFarlane, “Djot.” \href{https://djot.net/}{https://djot.net/} (accessed: Jan. 23, 2023).} \\{}
[2] & \leavevmode \hypertarget{bibentry:AsciiDoc}{Eclipse Foundation, “AsciiDoc.” \href{http://asciidoc.org/}{http://asciidoc.org/} (accessed: Jan. 23, 2023).} \\{}
[3] & \leavevmode \hypertarget{bibentry:mcilroy}{M. D. McIlroy, “A research unix reader: annotated excerpts from the programmer's manual, 1971-1986,” Bell Labs, 1987. Accessed: Mar. 21, 2023. [Online]. Available: \href{https://www.cs.dartmouth.edu/~doug/reader.pdf}{https://www.cs.dartmouth.edu/\textasciitilde{}doug/reader.pdf}} \\{}
\end{tabularx}
\renewcommand{\arraystretch}{1}
```

For the HTML output we could use either CSS grid or a table!

PR #325, opened 2023-05-13
Add styling for bib HTML and LaTeX output

adelhult:
Resolves GH-323. This PR adds enhanced 2 column styling for the bibliography when using the HTML or the LaTeX output format.
**Also, a bit unrelated:**  I ended up playing around with using a imports variable in the html template so this is included in the PR too.

**HTML**
Screenshot from 2023-05-13 07-42-51

**LaTeX**
Screenshot from 2023-05-12 22-20-02

**Plain text fallback**
Screenshot from 2023-05-12 22-19-07


packages/bibliography/src/main.rs
```diff
@@ -285,39 +289,133 @@ fn transform_bibliography(to: &str, input: &Value) {
...
+    match to {
+        // We want to use a tabularx environment when outputing latex where the prefix
+        // is in the first column and the rest in the second
+        "latex" => {
+            output.push(raw!("\n"));
+            output.push(import!("\\usepackage{tabularx}"));
+
+            output.push(raw!(
+                r"\renewcommand{\arraystretch}{1.5}
+\begin{tabularx}{\textwidth}{p{0.3cm} X}
```
Forsinge, line 340:
```suggestion
\begin{tabularx}{\textwidth}{p{0.3cm} >{\raggedright\arraybackslash}X}
```

Forsinge (commented):
Nice work! Would be nice to also avoid the horizontal stretching in each entry. I commented one way to achieve that.

adelhult:
Most of the issues pointed out should now be fixed. I only use a single column (or in the case of HTML a flexbox layout instead) when there is no prefix. The module can also handle corner cases where some entries have a prefix while others do not.
Screenshot from 2023-05-13 15-16-27

Screenshot from 2023-05-13 15-17-50

packages/bibliography/src/main.rs
```diff
@@ -286,38 +290,204 @@ fn transform_bibliography(to: &str, input: &Value) {
...
+            );
+
+            output.push(raw!(
+                r"\end{tabularx}
+\renewcommand{\arraystretch}{1}
+"
+            ));
+            output
+        }
+        // For html output we use a simple css grid layout
```
Forsinge, line 403: It looks like bibliographies without prefix use `display: flex` now, could update comment to reflect that

Forsinge (approved):
Seems like it works well now, good job!

packages/bibliography/src/main.rs
adelhult: thanks! Will fix!

PR #330, opened 2023-05-14
GH-210: Change label in LaTex to raw to avoid escaping

axoxelol1:
Simple fix, labels like `modmark_example` were getting converted to `modmark\_example` due to inline_content instead of raw.


CMDJojo (approved):
Lgtm!

Issue #340, opened 2023-05-15
Allow loading Gists in Playground


CMDJojo:
It would be nice to be able to load gists directly into the playground

PR #341, opened 2023-05-15
GH-340: Allow loading gists from the playground

CMDJojo:
Resolves GH-340.

This PR makes Playground check the query parameter `gist` on load, and if a gist ID is there, it will be loaded. So, if you go to [modmark.org/#/playground?gist=762632131a2726f19ca5f87e1f88b1be](https://modmark.org/pr-preview/pr-341/#/playground?gist=762632131a2726f19ca5f87e1f88b1be), it will load one an example gist I made. This is a nice way to share ModMark documents


bruhng (approved):
Smart idea with Gist

Issue #362, opened 2023-05-18
Smart punctuate crashes on escaping non-ascii character


CMDJojo:
**Describe the bug**
When escaping a non-ascii character with a backslash, the compiler crashes due to a panic in `punct.rs:try_smart_sequence` (probably due to the call to `seq.truncate`, see the `panic` section of its rustdoc).

**To Reproduce**
1. Try to compile the text `\😀`
2. See that, in the CLI, it crashes due to a panic, and in the Playground, it just says `Compiling...` and logging JSON errors in the console

PR #365, opened 2023-05-19
GH-362: Check for char boundary before truncating

CMDJojo:
This PR resolves GH-362 by checking that the index `try_smart_sequence` is going to split at is actually a char boundary. If not, it tries to backtrack not only one but two bytes, and if not, tries to backtrack three bytes etc.

You can try this out by doing `\😀` in the playground. Previously, the compiler would crash (and the status would be stuck at Compiling...), but now it works as expected. This is the code handling the smart sequences `...`, `--` and `---`, so you can try them out to see that they have not been broken. There might be better implementations of this, but I think this would be sufficient.

### DIFF
--- a/parser/src/punct.rs
+++ b/parser/src/punct.rs
@@ -145,7 +145,13 @@ fn try_smart_sequence(str: &mut String, last_escape: Option<usize>) {
             .take_while(|ch| ch == &last_char)
             .collect::<String>();
         if let Some(i) = last_escape {
-            seq.truncate(str.len() - i - 1);
+            for x in 1.. {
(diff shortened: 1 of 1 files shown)


adelhult:
Brakes on: `\k  ---` :(

CMDJojo:
No longer brake

adelhult (approved):

adelhult:
Looks good to me, but perhaps @Forsinge has more insight!

Forsinge:
Looks good to me as well! Maybe there is a more descriptive name for `x` but I can't think of anything clever.

Issue #368, opened 2023-05-19
Add system to manage document structure


Forsinge:
**Is your feature request related to a problem? Please describe.**
It would be nice to have a language independent system to manage numbering for headings, figures, tables as well as labels and references.

PR #369, opened 2023-05-19
Add structure system

Forsinge:
Resolves: #368 

This PR adds a system for managing table of contents + numbering for headings, figures and tables.

- To make normal headings numbered you can set `heading_style` to `numbered`.
- To make captions in figures and tables numbered you can set `caption_style` to `numbered`.

Example:
```
[config]
set heading_style numbered
set caption_style numbered
```

The structure package also provides several kinds of headings:
 - `[unnumbered-heading]`
 - `[numbered-heading]`
 - `[standalone-heading]`

Also included in the package:
 - `[table-of-contents]`
 - `[element-number]` (meant for internal use only)
 - `[label-to-key]` (meant for internal use only)


packages/structure/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+}
+
+pub struct StructureCounter {
+    headings: Vec<u64>,
+    figures: u64,
+    tables: u64,
+}
+
+#[derive(PartialEq)]
+enum TOCEntryType {
```
adelhult, line 22: I think its more common to still using PascalCase for abbreviations like TOC. But not that important! : )
```suggestion
enum TocEntryType {
```

packages/structure/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+        }
+    }
+
+    pub fn push_figure(&mut self) {
+        self.figures += 1;
+    }
+
+    pub fn push_table(&mut self) {
+        self.tables += 1;
+    }
```
adelhult, line 76: Might be nice to use a Iterator on a range instead  (but also not important!)

packages/structure/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+            children: vec![],
+            mode: Empty,
+        }
+    }
+}
+
+impl TOC {
+    // Build and return a TOC from the elements in "structure". Currently, only "numbered-heading"
+    // and "unnumbered-heading" are added.
+    pub fn build_from_list(max_level: usize) -> Self {
```
adelhult, line 129: Nice! This would be ease to generalize for a more flexible [listings] module in the future if users want to create a table of figures etc. too.

packages/structure/src/main.rs
```diff
@@ -0,0 +1,244 @@
...
+    print!(
+        "{}",
+        json!({
+            "name": "structure",
+            "version": "0.1",
+            "description": "This package manages document structure to provide numbering for headings, figures and tables.",
+            "transforms": [
+                {
+                    "from": "table-of-contents",
+                    "to": ["any"],
```
adelhult, line 158: Would be really nice to support all formats similiar to how the bibliography module works by generating other ModMark modules like link etc. instead of handling the low-level transformation.

But since we on line 124 will produce a error if the format is not HTML I think we should say just "html" in the manifest. Or add more general support (at least for LaTeX if we feel like we have time). 

```suggestion
                    "to": ["html"],
```

adelhult (commented):
Great job :smile:! I have tested it a bit and took a look at the source code too. It's quite a lot to digest but I think it looks good.

There are of course a lot of interesting higher level API decisions worth thinking to make this system more general. But I'm fine with leaving those issue for another time, in favour of getting something that works :)

PR #377, opened 2023-05-23
LaTeX support in plot package

adelhult:
Resolves GH-376 by adding latex support to the plot package. The PR also fixes two bugs: html output for captions and integer overflow (Carl must of missed it since only debug builds cause panic on integer overflow in Rust)


hugomardbrink (approved):
Nice work :)

